PR #5, opened 2019-01-28
Bugfix: Allow matrices with a single bus entry.

euronion:
Previously matrices with a single entry created a 0d numpy array, causing a ValueError in xr.DataArray(...) creation.
By reshaping instead of squeezing, a 1D array is enforced.

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -172,7 +172,7 @@ def convert_and_aggregate(cutout, convert_func, matrix=None,
     if per_unit or return_capacity:
         assert aggregate_func is aggregate_matrix, \
             "One of `matrix`, `shapes` and `layout` must be given for `per_unit`"
-        capacity = xr.DataArray(np.asarray(matrix.sum(axis=1)).squeeze(), [index])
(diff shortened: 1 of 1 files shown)


coroa:
`.ravel()` would be the ideal solution.

euronion:
Could you elaborate, why "ideal"?
I don't see any difference except for that it would guarantee a contiguous array.

Personally I find `.reshape(-1)` to be better readable and am a bit afraid of the `order=` keyword of `.ravel()`. ;-)

coroa:
Ideal as in explicitly created for this particular purpose to reinterpret a multidimensional array as one dimension, so it is more specific than `reshape`, and thus !could! be implemented in a faster way (for all I know it is not). The `order` keyword is only relevant if you have more than one dimension > 1 (not strictly true, if the array resulted from a slice with a non-unity step-size) and will not affect the outcome in this simple example at all.

I'd wager the readability issue for you is more of a familiarity difference. Anyway, let's stop bike-shedding. It does barely matter.

i'll merge as-is (spent too much time on it already),
Thanks

PR #7, opened 2019-02-07
Prevent warning from arcsin(..) and cosincidence comparison

euronion:
Machine eps causes values slightly below -1. and above +1. ; these
values result in NaN values from the following arcsin which in return
cause warning from comparison of the cosincidence matrix.
Clipping removes the additional +/- eps.


euronion:
Fair point.

Adjusted.

nworbmot:
Perfect, thanks for contributing this fix!

PR #8, opened 2019-02-10
Include comments on ERA5 dataset download

euronion:
Thank you for updating to the CDS API for ERA5! @coroa 

Since you kind of beat me to it, I figured I'll at least provide the comments
I added to the code while coding my own version.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -41,7 +41,7 @@ It can process the following weather data fields:
 * Temperature
 * Downward short-wave radiation
 * Upward short-wave radiation
-* Wind speeds
+* Wind 
 * Runoff
 * Surface roughness
 * Height maps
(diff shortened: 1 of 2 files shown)


coroa:
Thanks for the comments. I'd say I beat you proper and square to it :).

nworbmot:
Hi both, I was also about to do the CDS change before I saw @coroa had done it (I had already spent 1 hour looking at the interface). Maybe next time we should communicate better to avoid triple work? :-)

euronion:
> [...] Maybe next time we should communicate better to avoid triple work? :-)

Agreed! =)
Probably the easiest & most transparent way to go would be opening and (self-) assigning issues?

coroa:
I think that's the best option, yes!

PR #9, opened 2019-02-19
Introduce dataset caching and outsource wind speed extrapolation

euronion:
I introduced a simple dataset caching mechanism, which removes the overhead of recreating and loading netcdf file when repeatedly converting data to time-series.
The option can be activated using `cache_datasets=True` for the conversion functions when it is passed to `_convert_and_aggregate()` .

**Further changes:**

* Outsourced wind speed extrapolation to own method.\
  *New:* If more than one wind-speed is present in the dataset, it now chooses the wind speed at heights closest to the extrapolation target.
* Substituted `numpy.interp` by `xr.interp`: Function keeps coords, thus making the code slimmer (explicit xr.DataArray() conversion becomes obsolete)
* Introduced a `close_data()` method for clearing the newly introduced cache.
* Updated the README

**Gain**
Tested locally against my cutout for wind and PV - no changes are caused by this PR.
Speedups achieved for my rig: up to 4x (wind), up to 2x (PV)

**Caveat**
Uses Python3 `.format()`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,7 +77,7 @@ Getting started
 * Install atlite from this repository with all its library dependencies
 * Download one of the weather datasets listed above (ERA5 is downloaded
   automatically on-demand after the ECMWF
-  `cdsapi<https://cds.climate.copernicus.eu/api-how-to>` client is 
+  `cdsapi <https://cds.climate.copernicus.eu/api-how-to>` client is 
   properly installed)
 * Adjust the `atlite/config.py <atlite/config.py>`_ directory paths to
   point to the directory where you downloaded the dataset
@@ -87,6 +87,21 @@ Getting started
 * Select a sparse matrix of the geographical points inside the cutout
   you want to aggregate for your time series, and pass it to the
   appropriate converter function - see `examples/ <examples/>`_
+
+
+FAQ
+===
+
+* Which weather dataset should I use?
+
+  If you don't know which dataset to use, the ERA-5 dataset is an easy
+  and good way to start with.
+* I am doing repeated conversions, is there a way to make things go faster?
+
+  Some conversion function support and option via the keyword argument
+  `cache_datasets=True`. With this option, dataset objects are reused
+  and kept in memory. If you have enough memory or a small cutout,
+  this will probably make things faster for your
 
 Licence
 =======
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -32,6 +32,9 @@ import scipy as sp, scipy.sparse
 from six import string_types
 from operator import itemgetter
 
+import logging
+logger = logging.getLogger(__name__)
+
 from .aggregate import aggregate_sum, aggregate_matrix
 from .gis import spdiag, compute_indicatormatrix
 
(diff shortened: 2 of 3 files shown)


coroa:
Hi @euronion,

I am glad you Aarhus guys are joining the fray, it's a lot more lively all of a sudden and you are adding good stuff!

Two things: I have been putting off a major revision of how I am imagining atlite should work for some time now which will simplify the logic a lot. I am attacking this now, as in tonight and tomorrow night. Fingers crossed that it works out.

I'll only merge afterwards, hang tight.

In the mean time:
1. `convert.py` should stay small and tidy, so please move your new extrapolate windspeed function into a separate `wind` module as with f.ex. `hydro`!

2. it probably makes sense to split this pull request into two separate prs extrapolation and caching. and perhaps wait with the caching bit until tomorrow or the day after.

3. don't worry about `.format`. it has been around since `2.7`, and I even started using python3 only functionality in other parts of atlite already like the multiple context manager `with` in https://github.com/FRESNA/atlite/blob/7837f984a6dbb0c42a7adf91dc421835ab023bf0/atlite/datasets/era5.py#L191 .
We only need to keep python2 a bit alive for creating cutouts from `ncep`, as pynio does not work on 3, as of my last check. We should in a possibly never reached point in time port that dataset to `cfgrib` (http://www.matteodefelice.name/post/c3s-multimodel/ )

euronion:
Hi @coroa 

(OT: It's actually Giessen, I only got to enjoy the hospitality of the Aarhus for a couple of weeks in December :)

I'm looking forward to the new major revision.
After it is out, I'll implement your suggestions. I guess putting everything into one was just me being lazy and working on two things at the same time.

Regarding 3.:
Happy to hear that's the only reason we care about Python 2.7 .
I'm not familiar with `pynio` and have to admit I did not manage to get `ncep` running, but I briefly skimmed their website: [PyNIO V1.5.1 seems to be 3.x compatible](https://www.pyngl.ucar.edu/whatsnew.shtml#PyNIOVersion1.5.1) .

euronion:
Update:
While doing stuff™ I realised I hadn't `netCDF4` installed (as it is in the `environment.yaml` but not in the `setup.py`; so my script was falling back to the `scipy.io.netcdf` module.

With `netcdf4` installed, the **PR does not bring any speed benefits.**

I'll look into it, probably a lazy loading thing.

hailiangliu89:
> Hi @coroa
> 
> (OT: It's actually Giessen, I only got to enjoy the hospitality of the Aarhus for a couple of weeks in December :)
> 
> I'm looking forward to the new major revision.
> After it is out, I'll implement your suggestions. I guess putting everything into one was just me being lazy and working on two things at the same time.
> 
> Regarding 3.:
> Happy to hear that's the only reason we care about Python 2.7 .
> I'm not familiar with `pynio` and have to admit I did not manage to get `ncep` running, but I briefly skimmed their website: [PyNIO V1.5.1 seems to be 3.x compatible](https://www.pyngl.ucar.edu/whatsnew.shtml#PyNIOVersion1.5.1) .

`Pynio` works fine on Python 3 now. I have been using it for ncep data.

@coroa 3 `cfgrib` uses different variable names than pynio when reading `.grb2` files. That would be really annoying.

coroa:
Ok, then we'll just slowly let python2 support die with the next release.

PR #10, opened 2019-02-27
(Refractor) Wind speed extrapolation into new module

euronion:
Logically the two actions (wind speed extrapolation and conversion) are independent steps - this is represented by the separation.
Minor: In the same go, change from numpy.interp to xarray.interp

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -357,7 +357,7 @@ def convert_wind(ds, turbine):
     """Convert wind speeds for turbine to wind energy generation."""
 
     V, POW, hub_height, P = itemgetter('V', 'POW', 'hub_height', 'P')(turbine)
-    power_curve = xr.DataArray(POW, [('wind speed', V)], name='turbine power curve')
+    power_curve = xr.DataArray(POW/P, [('wind speed', V)], name='turbine power curve')
 
     wnd_hub = windm.extrapolate_wind_speed(ds, to_height=hub_height)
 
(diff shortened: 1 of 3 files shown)


coroa:
Thanks, great work!

euronion:
## Important: Outstanding problem

Previously `np.interp` was used, with `xarray.DataArray.interp` we actually get into trouble if we use the standard power curves (unsmoothed) provided by atlite or manufacturers:
* They are not strictly monotonic in `V`
* `numpy.interp` and `scipy` accept this (see https://github.com/scipy/scipy/issues/9886 and https://github.com/numpy/numpy/issues/10448#issuecomment-468215702 )
* `xarray.DataArray.interp` doesn't accept such a "function" and fails with an
```
ValueError: Expect x to be a 1-D sorted array_like.
```

I still favour the PR with the `xarray` solution, as it removes ambiguity with the power curve issue.


## Comments on additional commits
Two things I embarrasingly missed (and caused myself headaches with):
* https://github.com/FRESNA/atlite/pull/10/commits/8df5a22ec6a3ea9ba351c3bd0e75b9f535efd6fe fixes the missed rescaling of the max. turbine power
* https://github.com/FRESNA/atlite/pull/10/commits/0d4d63314591b9b844541f87aa56c86051903ef4 is obviously what one would expect from the function. Without, the PR does fail in some cases, as the `np.array(...)` conversion is removed in the refractor.

euronion:
> 
> 
> Thanks, great work!

Hey @coroa ,

Thanks :)

However: I was drilling down a problem and was slower than you.
Please check my comment above.

PR #12, opened 2019-03-01
Revert "(Refractor) Wind speed extrapolation into new module"

coroa:
Reverts FRESNA/atlite#10

Fails with standard wind turbines.


euronion:
I could provide the standard wind turbines profiles with a slightly increased cutoff velocites, e.g. `25+1e-6` instead of `25`.
Problem basically that the power curves are not well-defined at cutoff velocity, but have two values for the same point.

Or do an attempt to fix it during loading wind turbines (magic & not robust).

I think an unambigious power curve and failing with ambigious onces is better than silently continuing with `np.interp`.

coroa:
Puh, you move quickly and decidedly! If not always in the optimal direction. :).

I pushed a couple of quick fixes as ee5fbbeae53f0a10dc2d43460cd3c88400eb5062 and fa28dd2dd12c65f34bbb687e0117ac3dea8c5f79 .

coroa:
`np.interp` is fine with the dis-continuous wind turbines, we use in the default library. We will have to stick with that, until the support from xarray catches up, which is unlikely, since it is quite an edge case.

euronion:
> Puh, you move quickly and decidedly! If not always in the optimal direction. :).

Sorry, I am still trying to align myself with the optimal direction. Your remarks and code changes really help to get a feeling for it, thanks! :)

coroa:
Don't worry, I am glad for your high-quality contributions.

PR #13, opened 2019-03-01
(Fix / Sci) set rated power for smoothed turbines

euronion:
If we do not set the rated power after smoothing a turbine curve, the per unit calculation in convert_wind(..) will not be correct. Instead we would be introducing a 'hidden factor'.

(as described by
Norgaard, Per, and Hannele Holttinen. "A multi-turbine power curve approach." Nordic wind power conference. Vol. 1. Chalmers, 2004. )


coroa:
Good catch!

Issue #14, opened 2019-03-15
cyordereddict replacement


baerbock:
cyordereddict is obsolete! Since Python 3.5 it is replaced by built-in `collections.OrderedDict`
https://github.com/shoyer/cyordereddict

Is it still in use at atlite?

euronion:
No, it's just an artifact.
`collections.OrderedDict` has been used instead for quite some time now.

I'll do a quite PR

PR #15, opened 2019-03-17
Remove unused dependencies

euronion:
Removing some old and unused dependencies from `setup.py` and the conda environment.
Related to #14 .


coroa:
Thanks for that PR.

`bottleneck` is an optional dependency for `pandas` and `xarray` to speed up some rolling window operations, which we do use, so i'd like to keep it.

euronion:
Makes sense, also saw it is recommended by `xarray`.
But then do we also want to add [`numexpr`](https://github.com/pydata/numexpr) as recommended by `pandas` [here](https://github.com/pydata/numexpr)?

coroa:
I am not using anything where it would help, as far as i now, mainly the `df.eval` and `df.query` functions; unless someone recommends otherwise, i'd leave it out. Also the heavier calculations happen in `xarray` which does not have it as a dependency.

euronion:
It is recommended by `pandas` and is compatible with `xarray` and the like.
I just installed it and gave it a try (no import, just installation from `conda-forge`): speedup of ~15% (*).

It doesn't seem to hurt, I suggest we embrace it.

(* With my custom branch (implementing the dataset caching) - results should be applicable to this branch as well).

PR #16, opened 2019-04-23
Fix dependencies to include geopandas

euronion:
Geopandas became a necessary lib with the hydro module (7837f984a6dbb0c42a7adf91dc421835ab023bf0).

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -20,3 +20,4 @@ dependencies:
   - conda-forge::rasterio
   - conda-forge::shapely
   - conda-forge::libgdal
+  - conda-forge::geopandas
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ setup(
                       'dask>=0.18.0',
                       'rasterio',
                       'shapely',
-                      'progressbar2'],
+                      'progressbar2'
+                      'geopandas'],
     classifiers=[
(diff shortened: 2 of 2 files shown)


hailiangliu89:
May I add that Geopandas revolutionized the way we deal with geospatial
data. :)

Best regards,/顺颂时祺，

Hailiang/海亮

Hailiang Liu
Founder
Sente AI
CVR: 37509868
Tel: 91849404

coroa:
Sorry, I forgot! and thanks for the PR.

Side-note: Did you force-push this PR recently? I think, I noted you had removed the libgdal and one or two other dependencies and wanted to comment on that :).

euronion:
Force push? Atleast not intentionally, so I hope I didn't  ;-).

But I do not remember anything regarding the libgdal dependency in `setup.py`.
It only appears in the `environment.yaml`.

OT: Stumbled upon this line which would be happy to recieve an update ;-)
https://github.com/FRESNA/atlite/blob/1469dd8b7648938628f7c939930ddecc9d3058b9/setup.py#L14

coroa:
Hmm .. don't know where I saw it then. Some PR on pypsa-eur instead, presumably. (I'm actually still reading mail on that old address, but will update).

PR #20, opened 2019-06-12
Preparation of v0.2

coroa:
The first official version of atlite is getting closer and there is a bunch of good stuff coming. 

## Changes

The main change is that a cutout corresponds to a single netcdf file for the whole cutout period, which is fully accessible as a xarray at `cutout.data`.

1. This makes it possible to iterate over the data in customizable slices: `cutout.wind(shapes=countries, turbine="Vestas_V90_3MW")` uses months as in the previous version, while f.ex. `cutout.wind(..., windows='Y')` uses years. `windows` can be anything that `pd.Grouper` understands, ie. 'D', 'M', 'Y' or even '2D'. `windows = False` makes it possible to apply the conversion function to the data as a whole. It should also be possible to choose windows compatible with a particular time zone to avoid the re-averaging that was necessary for `heat_demand`.

2. The data for cutouts is now grouped into different `features` (from the ERA-5 dataset):
   ```python
   features = {
       'height': ['height'],
       'wind': ['wnd100m', 'roughness'],
       'influx': ['influx_toa', 'influx_direct', 'influx_diffuse', 'influx', 'albedo'],
       'temperature': ['temperature', 'soil_temperature'],
       'runoff': ['runoff']
   }
   ```

   It's possible to prepare a cutout only for a subset of the available features: `cutout.prepare(['runoff', 'wind'])`. One can always extend the cutout by running `prepare` again.

3. One can load a cutout fully into memory using `cutout.data.load()` (or `cutout.data.wnd100m.load()`, `cutout.data.roughness.load()` for wind only), which should fully supersede @euronion 's caching from #9 .

4. It's easy to get a subset of a cutout using `.sel` directly induced by atlite's `sel` function: `cutout.sel(time="2012-01")` or `cutout.sel(time="2012-07", bounds=german_shape.bounds)`

## Open questions

- [x] `config.py` has been completely removed instead one has to provide the necessary paths explicitly when creating new cutouts. In addition we could allow reading in a config file like `~/.atlite.config` or some such?
- [x] Should data cleaning methods be moved into datasets (ie surface roughness <= 0. ->0.002)? I think that would be a good idea! Are there objections?
- [x] When data is read in as dask arrays, it is not mutable in the conversion functions leading to Exceptions. We can either change everything to copy-on-change (ie use clip) or catch the error and throw a more helpful error message to have to prepare the dataset? Related to #30. To be conservative we load dask arrays before they are passed to the conversion functions.

## Remaining TODOS

- [X] Add a .sel method to produce a view on a subset of the data
- [ ] Mailing list for announcements
- [x] Release notes (done in `documentation` branch)
- [ ] Migration instructions
- [x] Examples should set up for showing warnings: #27 . (done in `documentation` branch)
- [ ] Other datasets:
  - [x] sarah
  - [ ] cordex
  - [ ] ncep
  - [ ] efas
- [x] Merge `documentation` branch

I'm happy about everyone, who wants to test the new version, provide feedback or help with documentation or the remaining todos! @leonsn @nworbmot @FabianHofmann @schlott @fneum


coroa:
@euronion : Do you have time to test whether this branch works for you?

The easiest invocation to get a cutout for wind generation covering germany now is something along the lines of:
```python
germanshapefile = ...
cutout = atlite.Cutout("<cutoutname>", bounds=germanshapefile.buffer(0.2).bounds, time="2012", module="era5")
cutout.prepare(["wind"])
```

This generates a `<cutoutname>.nc` file in the current directory containing wind speed and surface roughness in hourly resolution. This file is opened and fully accessible as `cutout.data`.

euronion:
I'd love to, but I can't spare sufficient ressource on it at the moment.
I estimate I can look at it in ~2 weeks earliest and then do some testing work:

I would also 
- [ ] include my config branch (https://github.com/euronion/atlite/tree/config-change) which allows for a config file.

And do I would at the same time work on an  
- [ ] update of the documentation with the official upcoming version.

including
- [ ] some examples to follow along.

I really hope you can wait that long :)

coroa:
@euronion We're not under any particular time pressure right now. I'd like to merge this around mid-July. Would be great if you can integrate your changes. I think it would be ideal to work on pull requests against this branch!

euronion:
@coroa Yes o/c only PR against this branch. Mid of July sounds good.

coroa:
Atlite in the current version `0.0.2` is now available from PyPI as well as conda-forge. The version prepared in this branch will be tagged as `0.1` (sic!) as soon as I merge this branch (the branch name will stay v0.2, for the time being).

euronion:
Hi @coroa ,
Other things on my side took longer than expected, I started working on the changes this week,
expect some PRs from me at the end of this week or start of next week :)

euronion:
Cross-checking:
Dropping support for Python 2.X is understandable, but was it intended to also drop support for <3.6?
I came across the use of f-strings which are not supported on lower Python versions.

In any case think best practice is to include the Python versino into the `setup.py`? E.g. with
```
python_requires='>=3.6'
```

coroa:
Semi-intended. I don't want to miss out on f-strings anymore, and even Debian stable these days already ships with python 3.7 so a requirement for 3.6 seems fair to me. Is there any reason to have a python 3.5 or lower?

Making the requirement explicit in `setup.py` and the conda-forge recipe should be done, true.

euronion:
Mainly just asking.
The only ad-hoc reason would be that you might want to apply the same restrictions on the Python version for `atlite` as for e.g. `pypsa`.

atlite/data.py
```diff
@@ -0,0 +1,215 @@
...
+        return [cutout.data]
+    else:
+        return Windows(cutout, features, window_type, allow_dask)
+
+class Windows(object):
+    def __init__(self, cutout, features, window_type=None, allow_dask=False):
+        group_kws = {}
+        if window_type is None:
+            group_kws['grouper'] = pd.Grouper(freq="M")
+        elif isinstance(window_type, base_string):
```
euronion:
```suggestion
        elif isinstance(window_type, str):
```

as `base_string` is no longer available in Python 3.X.

euronion:
The `windows=...` option does not return an error after changing `base_string` to `str` and seems to work as intended (good).
But I do not understand its purpose.

coroa:
`windows` is an argument available to any function wrapped by a `@requires_windowed` decorator. It allows you to choose how to traverse the `cutout.data` dataset.

Omitting it, is equivalent to the argument `windows = "M"`. This will break `cutout.data` into monthly chunks, so that the for-loop in `cutout.convert_and_aggregate` iterates over slices of a month. Internally, the `windows` argument is converted to an iterator by code similar to the following:
```python
windows = xr.core.groupby.DatasetGroupBy(self.data, self.data['time'], grouper=pd.Grouper(freq="M"))._iter_grouped()
```
over which the `convert_and_aggregate` function then iterates and calls the conversion functions.

Other allowed strings are `"2M"` for slices of two months, and `"D"` for days, `"Y"` for years. If you supply an integer it is used to feed the `bins` argument of `xr.core.groupby.DatasetGroupBy`. For instance `windows=2` splits the time axis in two. We will have to experiment with what good choices are.

This grouping mechanism can be switched off using `windows=False`. For a regular dataset as loaded from a cutout NetCDF file, xarray will then try to convert the whole data in one go. If your memory is big enough and you want to do a lot of repetitions it's probably best to preload the wind data
```python
for feature in cutout.dataset_module.features["wind"]:
    cutout.data[feature].load()
```
and convert it in one go
```python
cutout.wind(turbine=..., windows=False)
```

What needs to be investigated a bit further is the possibility to use dask automatically for the heavy lifting:
`cutout.data.chunk(time=744)` will split the dataset into approximately monthly chunks, and then `cutout.wind(turbine=..., windows=False)` will use dask to do the wind conversion for these monthly chunks. With the right configuration of dask this should even happen in parallel on multiple processors.

euronion:
What the `windows` argument is becoming somehow clear.
Still I do not see any benefits and reason for it.
Except for cosmetic changes (the resolution of the progressbar) and (maybe a slight) performance difference, it does not do much?

Preloading features has become really easy and convinient with the change.
`xarray` and `dask` also do a nice job and keeping parts of the `cutout.data` loaded automatically (not all of it though, so preloading still gives a few seconds of extra performance for repeated calculations).

As a universal and really simple solution for the `chunk` sizes, using the new `auto` keyword from `dask` could work.
I.e. what I (successfully) tried was
```
cutout.data.chunk({dim:'auto' for dim in cutout.data.dims})
```

I doubt we will find a better universal solution for the `chunk` sizes. This has been a long standing issue for `xarray` and `dask` and depends on the `cutout` size (spatial and temporal) as well as confiugration, hardware setup and use case.
But that should not come as a suprise, that is always an issue of optimisation of parallelisation.

E.g.:
I was playing around a bit with it today and the best I could do was 8s / iteration (without `dask`) and ~18s with `dask` and a `chunk={'time':2500}` and the auto-`chunk` above with ~11s.

coroa:
The `windows` machinery enabled switching out the data backend completely, while keeping compatibility with non-dask ready conversion functions.

coroa:
I'd be open to throwing it out in a separate PR in which we transition everything to `dask` and don't incur huge performance losses in the process.

euronion:
No, let's have the `windows` feature included in the upcoming version.
The idea is powerful and should be the way to go in the future.

Not knowing what went wrong back then with `dask`, I'd say we could try again. For this use-case `dask` does not always seem to be the most performant choice. As you say: if we try it, we should also clock and measure it.

euronion:
Ok, reasonable arguments for changing it back.

Still we need to document
1. How to use / not use the option; I find the [configuration doc](https://atlite.readthedocs.io/en/latest/configuration.html) a fitting location.
2. How to download the necessary GEBCO data.
3. Explain/link to a reference which explains why GEBCO is more accurate or compares GEBCO vs. ERA5.

I'll implement the code changes.
Can you take care of the documentation? Except for (1.) I do not have the necessary knowledge.

coroa:
Hi @fneum, @euronion,

thanks for taking over some of the responsibilities of doing the final steps of this release. And sorry about my sluggish communication.

# Urgent updates

1. I threw out the configuration system, as there were a couple of cyclic design decisions which kept me going in circles. We can try to readdress this in a future release, maybe we don't really need it, anyway.
2. Since ERA-5 and SARAH by default seem now to use a from negative latitude to positive latitude ordering, I tracked back on my earlier decision to enforce y=slice(north_lat, south_lat), which was causing confusions. From now on, it is `y=slice(min_y, max_y)`, which is somewhat simpler to communicate, anyway. *Needs doc update*.
3. The idea of a `cutout_dir` directory has been thrown out as well, since a cutout now corresponds one-to-one to a netcdf file (plus an optional pickle of the spatial index for speeding up indicatormatrix calculations).

Changes 2 and 3, mean that the main init function has been received another update, which still needs to be reflected in the docs. To instantiate a new cutout you now use:
```python
cutout = atlite.Cutout("western-europe.nc", time="2011-01",
                                  x=slice(...), y=slice(20, 40))
cutout.prepare()
```
name has been renamed to path and includes ".nc", there is no cutout_dir by default. I have fixed docs on this branch to reflect that.

I am not a hundred percent happy with that; `atlite.Cutout` is too complicated. Probably out-of-scope of this release. So don't bother.

##
- ERA-5 cutout generation is well tested, but should maybe be compared with previous version
- SARAH-2 worked in small tests, haven't tested with european yearly datasets, maybe will run into memory problems then, hopefully not
- Docs for SARAH-2 are missing; beware new versions have been released; one needs to download the full globe for the interested time period from their servers and extract the files beneath some directory, which you supply as `sarah_dir` argument to atlite.Cutout.
- Docs for GEBCO to fix the crude height map of ERA-5 are missing.

I will merge this branch into master this weekend and have a new branch for release notes and finishing touches, unless someone of you comes up with a convincing reason of why that is not a good idea. I am getting quite some confused emails about individual small problems, which are getting annoying.

atlite/gis.py
```diff
@@ -213,8 +239,8 @@ def regrid(ds, dimx, dimy, **kwargs):
     kwargs.update(dst_shape=dst_shape,
                   src_transform=src_transform,
                   dst_transform=dst_transform)
-    kwargs.setdefault("src_crs", 'longlat')
-    kwargs.setdefault("dst_crs", 'longlat')
+    kwargs.setdefault("src_crs", dict(init='EPSG:4326'))
+    kwargs.setdefault("dst_crs", dict(init='EPSG:4326'))
```
fneum: In `geopandas>=0.7.0` this should be (I think)
```suggestion
    kwargs.setdefault("src_crs", 'EPSG:4326')
    kwargs.setdefault("dst_crs", 'EPSG:4326')
```

fneum:
We should also
- [ ] add migration guide to the documentation

euronion:
> 
> 
> Hi @fneum, @euronion,
> 
> thanks for taking over some of the responsibilities of doing the final steps of this release. And sorry about my sluggish communication.
> # Urgent updates
> 
>     1. I threw out the configuration system, as there were a couple of cyclic design decisions which kept me going in circles. We can try to readdress this in a future release, maybe we don't really need it, anyway.
> 
>     2. Since ERA-5 and SARAH by default seem now to use a from negative latitude to positive latitude ordering, I tracked back on my earlier decision to enforce y=slice(north_lat, south_lat), which was causing confusions. From now on, it is `y=slice(min_y, max_y)`, which is somewhat simpler to communicate, anyway. _Needs doc update_.

Am I correct in blaming this change for a changing order in how the cutout data is stored, i.e. the order in `cutout.grid_coordinates()` changing?
If yes, just noting this breaks scripts relying on the order recieved from this and similar function / attributes, e.g. in PyPSA-EUR.
Still I believe this change is reasonable.

coroa:
> Am I correct in blaming this change for a changing order in how the cutout data is stored, i.e. the order in `cutout.grid_coordinates()` changing?
> If yes, just noting this breaks scripts relying on the order recieved from this and similar function / attributes, e.g. in PyPSA-EUR.

Yes, `cutout.grid_coordinates()` changes its order due to this change.

Can you point me to code that breaks? Maybe there is a simple workaround?

euronion:
[OT]

The lines I am currently struggling with are https://github.com/PyPSA/pypsa-eur/blob/fbbf1ba0ee190ea5c9f2b88d5ad2559662d3e478/scripts/build_renewable_profiles.py#L328-L336 .
The order of `cell_coords` does not seem match the order of `layoutmatrix`.

coroa:
> The lines I am currently struggling with are https://github.com/PyPSA/pypsa-eur/blob/fbbf1ba0ee190ea5c9f2b88d5ad2559662d3e478/scripts/build_renewable_profiles.py#L328-L336 .

You probably only need to swap what you get from geokit/GDAL, ie. https://github.com/PyPSA/pypsa-eur/blob/fbbf1ba0ee190ea5c9f2b88d5ad2559662d3e478/scripts/build_renewable_profiles.py#L241 should become
```python
    return csr_matrix(gk.raster.extractMatrix(availability)[::-1].flatten() / 100.)
```

euronion:
Turns out it is not such a small change. I took me nearly a whole day to figure out, that this also affects the line on how cells are aggregated in
https://github.com/PyPSA/atlite/blob/7d60f68e9043cecd7f65c77b7e0f480c1f22d4b0/atlite/aggregate.py#L31

as the `y` dimension is now reversed.
I.e. it matters if the internal structure from before was assumed in scripts (like PyPSA-EUR, where I am fixing this now). I don't know if it matters for regular users though...

coroa:
> Turns out it is not such a small change. I took me nearly a whole day to figure out, that this also affects the line on how cells are aggregated in
> 
> https://github.com/PyPSA/atlite/blob/7d60f68e9043cecd7f65c77b7e0f480c1f22d4b0/atlite/aggregate.py#L31
> 
> as the `y` dimension is now reversed.
> I.e. it matters if the internal structure from before was assumed in scripts (like PyPSA-EUR, where I am fixing this now). I don't know if it matters for regular users though...

Can you elaborate, what exactly you think will have to be changed in PyPSA-Eur and ideally link to a PR where you do that, then we can discuss in a given place based on the actual code. I might of course be missing things, but I fail to see anywhere else, where the switch should not happen transparently (ie data changes, yes, but consistently and w/o code changes).

The reason why I think the switch in the ordering should cancel is because most code refers to `cutout.grid_coordinates` and `cutout.grid_cells`, which also changed its ordering accordingly, except for the averaging of the profiles from GLAES (hence my argument above).

euronion:
## Order or `grid_cells` and `grid_coordinates()`

I am thinking more along the line of: Are people expecting the order in the lists returned by `cutout.grid_cells` and `cutout.grid_coordinates()` to be reliable or not.
Maybe it does cancel out in `PyPSA-EUR` or one codes it independent of this order (I just did so).
But maybe people are expecting the order to be consistent between version, i.e. then we are changing the API behaviour in an unexpected way -> we should at lteast add this to the release notes.

## The bigger problem

The code with a more hideous and less obvious problem I am refering to is this one [1]:

```
    profile, capacities = func(
        # Redindex, stack index and conv. to CSR to be consistent with the format expected by atlite
        matrix=layoutmatrix.reindex_like(
            cutout.data).stack(spatial=('y', 'x')).data.tocsr(),
        index=buses,
        per_unit=True,
        return_capacity=True,
        show_progress='Compute profiles: ',
        **resource)
```


More specifically how the `matrix` passed to `atlite` has to be structured. This matrix is 2D, where one dimension is reserved for the `index=buses` and the other dimension is a `.stack()`-ed version of a 2D array.
Now the behaviour of `.stack()` depends on the order of the index, that's why the `.reindex_like(...)` is relevant here. In the version before it looked like this [2]:

```
    profile, capacities = func(matrix=layoutmatrix, index=buses, per_unit=True,
                               return_capacity=True, show_progress='Compute profiles: ',
                               **resource)
```

where the `matrix=layoutmatrix` as implicitly assuming the 2D array of values refering to the coordinates in the cutout to be stacked with descending `y` and ascending `x`. That was the old `atlite` structure.
If I use the same code with the new `atlite=v0.2` the result is different, because we assume (inside `atlite`!) that this matrix is stacked with ascending `y` and `x`.

It boils down to 
```
cutout.sortby('y', ascending=True).stack(spatial=('y','x'))
```
not being the same as
```
cutout.sortby('y', ascending=False).stack(spatial=('y','x'))
```
but we use
```
cutout.stack(spatial=('y','x')
```
here
https://github.com/PyPSA/atlite/blob/7d60f68e9043cecd7f65c77b7e0f480c1f22d4b0/atlite/aggregate.py#L31


## Possible solution

The easiest solution could be to guarantee the old behaviour by fixing the index order of created cutouts during their creation
```
cutout.data = cutout.data.sortby('y', ascending=False)
cutout.data = cutout.data.sortby('x', ascending=True)
```
This would allow us to make the change in the `.Cutout(...)` signature while ensuring backwards compatability.

## Links

[1] Source: New code in `PyPSA-EUR` for the upcoming `atlite` version
https://github.com/PyPSA/pypsa-eur/blob/283042d1dd3cd11b5313479c7ca03cbe58f778c9/scripts/build_renewable_profiles.py#L390-L398
[2] Source: https://github.com/PyPSA/pypsa-eur/blob/bb3477cd693d1c8e77e75e61a1a7e1a4647e6a3c/scripts/build_renewable_profiles.py#L303-L305

euronion:
Different topic:
I failed to run `cutout.pv(...)` for the last days and found the problem to be insufficient memory (the script ran slowly and crashed at one point in time without a tractable error).
I solved it with `windows="D"`, which requires less data to be loaded into memory at the same time.

@coroa : What do you think about making `windows="D"` default for the convert operations?

euronion:
I checked whether GEBCO is working and can confirm GEBCO 2014 and the new GEBCO 2019 are working. I added the necessary documentation and a helpful comparison.

atlite/cutout.py
```diff
@@ -1,120 +1,177 @@
...
+        x : (opt.) slice
+            Outer longitudinal bounds for the cutout (west, east).
+        y : (opt.) slice
+            Outer latitudinal bounds for the cutout (south, north).
+        time : (opt.) slice
+
+        """
+        if cutout_dir is not None:
+            warn("The argument `cutout_dir` has been deprecated in favour of just `path`", DeprecationWarning)
+            path = Path(cutout_dir) / path
```
FabianHofmann, line 82:
```suggestion
            path = Path(cutout_dir) 
```
This raised an error with my old code

euronion: Could you provide the error raised + `atlite.Cutout(...)` call signature you used?

FabianHofmann: Yes of course, I think I understand why you're hesitating. In my case `cutout_dir`  was a string and path was None (default). Then it tries to append `None` to the `cutout_dir` which is not working:
```python
  File "/home/fabian/vres/py/atlite/atlite/cutout.py", line 82, in __init__
    path = Path(cutout_dir) / path

  File "/home/fabian/.miniconda3/lib/python3.7/pathlib.py", line 920, in __truediv__
    return self._make_child((key,))

  File "/home/fabian/.miniconda3/lib/python3.7/pathlib.py", line 699, in _make_child
    drv, root, parts = self._parse_args(args)

  File "/home/fabian/.miniconda3/lib/python3.7/pathlib.py", line 653, in _parse_args
    a = os.fspath(a)

TypeError: expected str, bytes or os.PathLike object, not NoneType
```
The intention of this line is probably to select a nc file within the `cutout_dir`?

euronion: Probably to keep it backwards compatible to when we were using `cutout_dir´ and `name` rather than `path`.

I.e. this should be `name`, not `path`.

euronion:
@FabianHofmann Sorting should happen at the previous location before the
```
if data is None:
```

else loaded cutouts and freshly prepared cutouts might have a different order of their coordinates.

atlite/cutout.py
euronion: This should  now be fixed.

FabianHofmann:
Can you point out where the loaded data would be ordered according to the ordered slices? When sorting x and y at the previous location, the standard loading of a built cutout wouldn't work.

```python
era5 = Cutout(path="era5_test")
```
As the x and y are not defined.

euronion:
> Can you point out where the loaded data would be ordered according to the ordered slices?
> 
> The problem with the precious location: When sorting x and y at this location, the standard loading of a built cutout wouldn't work.

You are correct. `cutoutparams` is only used for uncreated cutouts. If cutouts are loaded, the attribute isn't used anyways and the representation for `print(cutout)` is constructed on the fly from the `cutout` indices.

-> NVM, no change necessary.

atlite/cutout.py
```diff
@@ -1,120 +1,196 @@
...
+            containing (min.long, min.lat, max.long, max.lat).
+        x : (opt.) slice
+            Outer longitudinal bounds for the cutout (west, east).
+        y : (opt.) slice
+            Outer latitudinal bounds for the cutout (south, north).
+        time : (opt.) slice
+
+        """
+        if cutout_dir or name:
+            warn("The arguments `cutout_dir` and `name` have been deprecated in
```
FabianHofmann, line 81: Here is a " missing a the end of the line, I'm pushing a small fixup

FabianHofmann: And I also see that it should be a line before
``` 
name = cutoutparams.get("name", None)
```
as it's not defined otherwise

euronion: No. Also my change and comment were wrong and corona's code was correct.

`name` is currently (`master` branch) the first argument in the `cutout` constructor. in `v0.2` it is superseded by `path`. It has never been in the `cutoutparams` (and should never be).

I'll bring everything together.

This PR is bloating, we should wrap it up and believe in our previous changes ;-)

FabianHofmann: I agree, it's getting really messy. But keep in mind the 'division' by path does not work if it is None.

euronion: Could you check if this version is now working for you?

FabianHofmann: Yes, works fine! It's also tested in the test script `test_cututout_preparation.py` in #71

euronion:
> 
> ## Order or `grid_cells` and `grid_coordinates()`
> 
> I am thinking more along the line of: Are people expecting the order in the lists returned by `cutout.grid_cells` and `cutout.grid_coordinates()` to be reliable or not.
> ...

Following the discussions with @FabianHofmann we will go (if no objections come up) with the following approach (which needs to be documented, i.e. let this serve as the reminder for the release notes):

- [ ] people should use the atlite method for creating an indicator matrix (we will link to an example / pypsa-eur script showing how it is done)
- [ ] people will be notified that the ordering changed
- [ ] functions now returning a different order of elements due to this will be mentioned in the release notes (and maybe we will change the return type from `list` to `set` .

coroa:
> Following the discussions with @FabianHofmann we will go (if no objections come up) with the following approach (which needs to be documented, i.e. let this serve as the reminder for the release notes):
> 
>     *



>     *



> Following the discussions with @FabianHofmann we will go (if no objections come up) with the following approach (which needs to be documented, i.e. let this serve as the reminder for the release notes):
> 
> 1. people should use the atlite method for creating an indicator matrix (we will link to an example / pypsa-eur script showing how it is done)
> 2. people will be notified that the ordering changed
> 3. functions now returning a different order of elements due to this will be mentioned in the release notes (and maybe we will change the return type from list to set .

**I like point 2 and the first half of 3**. Yes, please add a fat warning about this breaking change.

**About 1.**: GLAES uses rasters to track available areas, it is unable to produce shapes, which you would need to feed to `compute_indicatormatrix` and even if you find a way to create shapes, it would be very round-about (and probably lossy) to *take a raster* -> *create a shape* -> *create a coarser raster*. 

In an ideal world w/o time constraints, the possibility to create a GLAES-adjusted indicatormatrix would be abstracted from the `build_renewable_potentials` script into a new set of functions in `atlite.gis`. Imagine passing a user-written callback function to `convert_and_aggregate`, which receives an ExclusionCalculator object (the `ec` object) that it can operate freely and is then used to build the indicatormatrix automatically by atlite.

**NO set**: it is important that `grid_cells` and `grid_coordinates` maintain the order you get from stacking `(y, x)`. A set would break the atlite computed indicatormatrix and be very hard to predict correctly.

atlite/cutout.py
```diff
@@ -123,43 +233,83 @@ def shape(self):
...
+        return self.prepared_features == self.available_features
 
+    @property
+    def prepared_features(self):
+        features = atleast_1d(self.data.attrs.get("prepared_features", []))
+        return self.available_features.loc[:, features].drop_duplicates()
 
     def grid_coordinates(self):
+        logger.warning("The order of elements returned by `grid_coordinates` changed. "
+                       "Check the output of your workflow for correctness.")
```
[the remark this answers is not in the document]
euronion: So when I said "people will be notified about the changes" I meant that literally: People will get an active warning about this change when using the affected function/attribute.
This is a breaking change which should have been announced (in an ideal world) with a `DeprecationWarning` or similar beforehand. We don't have this luxury.

Warnings can easily be disabled. We can also remove it for the version after `v0.2`. But we are doing the same for (non-) automatic migration from the old to the new cutout format.
Not sure how this is any different.

If you want to change it go ahead - I won't undo it. I however feel throwing a warning is appropriate here.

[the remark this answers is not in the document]
FabianHofmann: Can there be something implemented as a one-time warning? Perhaps at the first call of atlite v0.2?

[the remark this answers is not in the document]
FabianHofmann: Just noticing we could have at least add the deprecation warning in v0.0.3... too little too late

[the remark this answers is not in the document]
euronion: > Can there be something implemented as a one-time warning? Perhaps at the first call of atlite v0.2?

I think logging a warning only once per logger on the first call after loading the module (i.e. once per script execution) is the default behaviour.

FabianHofmann:
finally :)

Issue #21, opened 2019-06-13
Add adaptor for loading windturbine powercurves from OpenEnergyDatabase


coroa:
RLI built a list of 139 recent wind turbines collected from 11 different sources. They re-published them as a [table on their oedb](https://openenergy-platform.org/dataedit/view/supply/turbine_library) under an ODbL.

We should provide a small adaptor and instructions for querying this table and using it in atlite.

[saio](https://github.com/coroa/saio) is a great help for that.

euronion:
Closing: #24 realises this feature.

The feature will be available in the new version which is currently in the last stages of development (see #20 ).

PR #24, opened 2019-07-22
Provide oedb adapter for downlading wind turbine configs.

euronion:
As stated in #21 , this change provides an adapter for downloading into `resources/windturbine` (more precisely `atlite.config.windturbine_dir`) and using the data from OEDB.

```
def download_turbineconf(requested_turbine=None):
    """Download a windturbine configuration from the OEDB database.
    
    Download the configuration of a windturbine model from the OEDB database
    into the local 'windturbine_dir'.
    The OEDB database can be viewed here:
    https://openenergy-platform.org/dataedit/view/supply/turbine_library
    (2019-07-22)
    Only one turbine configuration is downloaded at a time, if the 
    search parameters yield an ambigious result, not data is downloaded.
    
    Parameters
    ----------
    requested_turbine : dict
        Search parameters, either provide the turbine model id (takes priority)
        or a manufacturer and turbine name, e.g. the following are identical:
        {'id':10}, {'id':10, 'manufacturer':'Unknown', name:'E-53/800'},
        {'manufacturer':'Enercon', name:'E-53/800'}
        
    Returns
    -------
    turbineconf : dict
        The turbine configuration downloaded and stored is also returned as a dict.
        Has the same format as returned by 'atlite.ressource.get_turbineconf(name)'.
    """
```

*Depends on minor changes introduced in #23 .*

### DIFF
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -29,39 +29,37 @@ from operator import itemgetter
 import numpy as np
 from scipy.signal import fftconvolve
 from pkg_resources import resource_stream
+import ast
 
 from .wind import download_turbineconf
 
 import logging
 logger = logging.getLogger(name=__name__)
 
-def get_windturbineconfig(turbine, source='local'):
+def get_windturbineconfig(turbine):
     """Load the wind 'turbine' configuration.
 
     The configuration can either be one from local storage, then 'turbine' is
     considered part of the file base name '<turbine>.yaml' in config.windturbine_dir.
-    Alternatively it can refer to the unique 'id' of a turbine in OEDB, see
-    atlite.wind.download_turbineconf() for details.
+    Alternatively 'turbine' is the string representation of a dict refer to a unique
+    turbine OEDB.
 
     Parameter
     ---------
     turbine : str
-        Name of the local turbine file in config.windturbine_dir or a string
-        refering to a specific turbine online in the oedb with format 'oedb:<id>'.
-    source : str
-        (Default: 'local') Either 'local' or 'oedb'. If local, the 'turbine' is
-        considered the filename of a local turbine in config.windturbine_dir,
-        the file suffix (e.g. '.yaml') is not needed.
-        If 'oedb', then only the online database will be searched for a turbine
-        with 'id' provided in 'turbine'.
+        Name of the local turbine file (without .yaml suffix) in config.windturbine_dir
+        or a string starting with "oedb:" for an online lookup in the OEDB database.
+        The online database will be searched for a turbine according to the dict following
+        the flag, e.g. "oedb:{'id':'10'}".
+        For all supported fields, see atlite.wind.download_turbineconf() for details.
     """
 
-    if turbine.startswith('oedb'):
-        source = 'oedb'
-        turbine = turbine[5:] # Extract turbine id
+    if turbine.startswith('oedb:'):
+        source = 'oedb:'
+        turbine = turbine[5:] # Remove 'oedb:' switch
+        turbine = ast.literal_eval(turbine)
 
-    if source is 'oedb':
-        return download_turbineconf({'id':turbine}, store_locally=False)
+        return download_turbineconf(turbine, store_locally=False)
     elif source is "local":
         res_name = os.path.join(config.windturbine_dir, turbine+".yaml")
         res_name = construct_filepath(res_name)
@@ -69,9 +67,6 @@ def get_windturbineconfig(turbine, source='local'):
         turbineconf = yaml.safe_load(resource_stream(__name__, res_name))
         V, POW, hub_height = itemgetter('V', 'POW', 'HUB_HEIGHT')(turbineconf)
         return dict(V=np.array(V), POW=np.array(POW), hub_height=hub_height, P=np.max(POW))
-    else:
-        logging.info(f"Unknown option used for 'source':{source}.")
-        return None
 
 def get_solarpanelconfig(panel):
     res_name = "resources/solarpanel/" + panel + ".yaml"
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -31,6 +31,7 @@ from scipy.signal import fftconvolve
 from pkg_resources import resource_stream
 import ast
 
+from . import config
 from .wind import download_turbineconf
 
 import logging
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -37,7 +37,7 @@ from .wind import download_turbineconf
 import logging
 logger = logging.getLogger(name=__name__)
 
-def get_windturbineconfig(turbine, source='local'):
+def get_windturbineconfig(turbine):
     """Load the wind 'turbine' configuration.
 
     The configuration can either be one from local storage, then 'turbine' is
@@ -48,25 +48,35 @@ def get_windturbineconfig(turbine, source='local'):
     Parameter
     ---------
     turbine : str|dict
-        Name of the local turbine file for using source='local'.
-        For selecting a turbine from source='oedb', the dictionary is used to retrieve the
-        for a matching turbine, see atlite.wind.download_turbineconf() for details.
-    source : 'local' (default) |'oedb'
-        Location from where to retrieve the turbine configuration.
+        Name of the local turbine file.
+        Alternatively a dict for selecting a turbine from the Open Energy Database,
+        in thins case the key 'source' should be contained. For all other key arguments
+        to retrieve the matching turbine, see atlite.wind.download_turbineconf() for details.
     """
 
-    if source.lower() == 'oedb':
+    if isinstance(turbine, str):
+        if turbine.startswith('oedb:'):
+            turbineconf = download_turbineconf(turbine, store_locally=False)
+        else:
+            turbine = {'filename':turbine, 'source':'local'}
+    
+    if turbine.get('source') is None:
+        logger.warning("No key 'source':'oedb' provided with the turbine dictionary."
+                       "I am assuming and adding it for now, but still nag you about it.")
+        turbine['source'] = 'oedb'
+    
+    if turbine['source'] == 'oedb':
         turbineconf = download_turbineconf(turbine, store_locally=False)
-    elif source == "local":
-        res_name = os.path.join(config.windturbine_dir, turbine+".yaml")
+    elif turbine['source'] == "local":
+        res_name = os.path.join(config.windturbine_dir, turbine['filename']+".yaml")
         res_name = construct_filepath(res_name)
 
         turbineconf = yaml.safe_load(resource_stream(__name__, res_name))
     else:
-        raise Exception("Not a valid 'source'.")
+        raise ValueError("Not a valid 'source'.")
     
     if turbineconf is None:
-        raise Exception("No matching turbine configuration found.")
+        raise ValueError("No matching turbine configuration found.")
     
     V, POW, hub_height = itemgetter('V', 'POW', 'HUB_HEIGHT')(turbineconf)
     return dict(V=np.array(V), POW=np.array(POW), hub_height=hub_height, P=np.max(POW))
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -30,16 +30,48 @@ import numpy as np
 from scipy.signal import fftconvolve
 from pkg_resources import resource_stream
 
+from .wind import download_turbineconf
+
 import logging
 logger = logging.getLogger(name=__name__)
 
-def get_windturbineconfig(turbine):
-    """Load the 'turbine'.yaml file from local disk and provide a turbine dict."""
+def get_windturbineconfig(turbine, source='local'):
+    """Load the wind 'turbine' configuration.
 
-    res_name = "resources/windturbine/" + turbine + ".yaml"
-    turbineconf = yaml.safe_load(resource_stream(__name__, res_name))
-    V, POW, hub_height = itemgetter('V', 'POW', 'HUB_HEIGHT')(turbineconf)
-    return dict(V=np.array(V), POW=np.array(POW), hub_height=hub_height, P=np.max(POW))
+    The configuration can either be one from local storage, then 'turbine' is
+    considered part of the file base name '<turbine>.yaml' in config.windturbine_dir.
+    Alternatively it can refer to the unique 'id' of a turbine in OEDB, see
+    atlite.wind.download_turbineconf() for details.
+
+    Parameter
+    ---------
+    turbine : str
+        Name of the local turbine file in config.windturbine_dir or a string
+        refering to a specific turbine online in the oedb with format 'oedb:<id>'.
+    source : str
+        (Default: 'local') Either 'local' or 'oedb'. If local, the 'turbine' is
+        considered the filename of a local turbine in config.windturbine_dir,
+        the file suffix (e.g. '.yaml') is not needed.
+        If 'oedb', then only the online database will be searched for a turbine
+        with 'id' provided in 'turbine'.
+    """
+
+    if turbine.startswith('oedb'):
+        source = 'oedb'
+        turbine = turbine[5:] # Extract turbine id
+
+    if source is 'oedb':
+        return download_turbineconf({'id':turbine}, store_locally=False)
+    elif source is "local":
+        res_name = os.path.join(config.windturbine_dir, turbine+".yaml")
+        res_name = construct_filepath(res_name)
+
+        turbineconf = yaml.safe_load(resource_stream(__name__, res_name))
+        V, POW, hub_height = itemgetter('V', 'POW', 'HUB_HEIGHT')(turbineconf)
+        return dict(V=np.array(V), POW=np.array(POW), hub_height=hub_height, P=np.max(POW))
+    else:
+        logging.info(f"Unknown option used for 'source':{source}.")
+        return None
 
 def get_solarpanelconfig(panel):
     res_name = "resources/solarpanel/" + panel + ".yaml"
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -37,37 +37,39 @@ from .wind import download_turbineconf
 import logging
 logger = logging.getLogger(name=__name__)
 
-def get_windturbineconfig(turbine):
+def get_windturbineconfig(turbine, source='local'):
     """Load the wind 'turbine' configuration.
 
     The configuration can either be one from local storage, then 'turbine' is
     considered part of the file base name '<turbine>.yaml' in config.windturbine_dir.
-    Alternatively 'turbine' is the string representation of a dict refer to a unique
-    turbine OEDB.
+    Alternatively the configuration can be downloaded from the Open Energy Database (OEDB),
+    in which case 'turbine' is a dictionary used for selecting a turbine from the database.
 
     Parameter
     ---------
-    turbine : str
-        Name of the local turbine file (without .yaml suffix) in config.windturbine_dir
-        or a string starting with "oedb:" for an online lookup in the OEDB database.
-        The online database will be searched for a turbine according to the dict following
-        the flag, e.g. "oedb:{'id':'10'}".
-        For all supported fields, see atlite.wind.download_turbineconf() for details.
+    turbine : str|dict
+        Name of the local turbine file for using source='local'.
+        For selecting a turbine from source='oedb', the dictionary is used to retrieve the
+        for a matching turbine, see atlite.wind.download_turbineconf() for details.
+    source : 'local' (default) |'oedb'
+        Location from where to retrieve the turbine configuration.
     """
 
-    if turbine.startswith('oedb:'):
-        source = 'oedb:'
-        turbine = turbine[5:] # Remove 'oedb:' switch
-        turbine = ast.literal_eval(turbine)
-
-        return download_turbineconf(turbine, store_locally=False)
-    elif source is "local":
+    if source.lower() == 'oedb':
+        turbineconf = download_turbineconf(turbine, store_locally=False)
+    elif source == "local":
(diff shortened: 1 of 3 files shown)


coroa:
Great, thanks!, will review after #23. Would be awesome it it could be used on-demand, by requesting a turbine like `turbine="oedb:<oedb-name>"` in cutout.wind.

euronion:
Aaand that should now also be possible. + made some minor adjustments.

I wish the turbines in the OEDB would be better documented and described (meta data, consistent format). Right now, we can only hope to access a unique entry using {'id':<id>} or {'manufacturer':<m>, 'name':<n>}.

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+        OEDB_URL = 'https://openenergy-platform.org/api/v0/schema/supply/tables/turbine_library/rows'
+        result = requests.get(OEDB_URL)
+    except requests.exceptions.RequestException as e:
+        logger.info(f"Connection to OEDB failed with:\n\n{str(e)}")
+        return None
+
+    # Convert JSON to dataframe for easier filtering
+    # Only consider turbines with power curves available
+    df = pd.DataFrame.from_dict(result.json())
+    df = df[df.has_power_curve]
```
coroa: Let's store this DataFrame somewhere in the `atlite` module to avoid hammering the db; ie. `_oedb_turbines = None`, on the module level.

Here
```python
global _oedb_turbines

if _oedb_turbines is None:
    [... download ...]

File: atlite/wind.py [hunk 2]
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+    into the local 'windturbine_dir'.
+    The OEDB database can be viewed here:
+    https://openenergy-platform.org/dataedit/view/supply/turbine_library
+    (2019-07-22)
+    Only one turbine configuration is downloaded at a time, if the 
+    search parameters yield an ambigious result, not data is downloaded.
+    
+    Parameters
+    ----------
+    turbine : dict
```
coroa:
```suggestion
    turbine : dict|str
```

Let's also allow a *smart* string:
```python
if isinstance(turbine, str):
    try:
        turbine = {'id': int(turbine)}
    except ValueError:
        turbine = {'name': turbine}
```

this would only fail if the name is not specific enough that one has to specify a manufacturer, as well; and work fine otherwise.

atlite/resource.py
```diff
@@ -29,17 +29,45 @@
...
+        the flag, e.g. "oedb:{'id':'10'}".
+        For all supported fields, see atlite.wind.download_turbineconf() for details.
+    """
+
+    if turbine.startswith('oedb:'):
+        source = 'oedb:'
+        turbine = turbine[5:] # Remove 'oedb:' switch
+        turbine = ast.literal_eval(turbine)
+
+        return download_turbineconf(turbine, store_locally=False)
```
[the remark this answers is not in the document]
coroa: Either `download_turbineconf` should raise an error, or we have to raise an error if we encounter `None` here. Otherwise, `cutout.wind` will suddenly return a layout rather than a time-series, which is definitely unwanted!

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+        "name": ds['name'].strip(),
+        "manufacturer": ds.manufacturer.strip(),
+        "source": f"Original: {ds.source}. Via OEDB {OEDB_URL}",
+        "HUB_HEIGHT": ds.hub_height.strip(),
+        "V": json.loads(ds.power_curve_wind_speeds),
+        "POW": power.tolist(),
+    }
+
+    if store_locally is True:
+        filename = "{m}_{n}.yaml".format(m=turbineconf['manufacturer'],
```
coroa:
```suggestion
        filename = ("{manufacturer}_{name}.yaml".format(**turbineconf)
                    .replace("/", "_").replace(" ", "_"))
```

coroa: Still haven't figured out how to implement multiline suggestions :/

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+            
+            logger.error("'name' and 'manufacturer' must be provided. "
+                         "Alternatively provide a turbine id from OEDB.")
+            return None
+        logger.info(f"Searching turbine power curve in OEDB database using "
+                    f"manufacturer '{turbine['manufacturer']}'"
+                    f"and name '{turbine['name']}'.")
+            
+        ds = df[df.name == turbine['name']]
+        ds = ds[ds.manufacturer == turbine['manufacturer']]
```
coroa:
```python
ds = df[(df.name == turbine['name']) & (df.manufacturer == turbine['manufacturer'])]
```

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+
+    if len(ds) < 1 :
+        logger.info(f"No turbine found for {turbine}.")
+        return None
+    elif len(ds) > 2 :
+        logger.info(f"Provided information corresponds to more than one turbine. "
+                    f"Provide id to unambiguous lookup. {turbine}.")
```
coroa: We should show `ds` in the logging message, so that the user can choose.

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+        filename = filename.replace(" ","_")
+        filepath = utils.construct_filepath(os.path.join(config.windturbine_dir, filename))
+
+        with open(filepath, 'w') as turbine_file:
+            yaml.dump(turbineconf, turbine_file)
+            
+        logger.info(f"Turbine configuration downloaded to '{filepath}'.")
+
+    if ";" in turbineconf['HUB_HEIGHT']:
+        logger.info("Multiple HUB_HEIGHTS in dataset. Manual clean-up required.")
```
coroa:
```suggestion
        logger.warn(f"Multiple HUB_HEIGHTS in dataset ({turbineconf['HUB_HEIGHT']}). Manual clean-up required.")
```

atlite/wind.py
```diff
@@ -90,3 +101,105 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+    """Download a windturbine configuration from the OEDB database.
+    
+    Download the configuration of a windturbine model from the OEDB database
+    into the local 'windturbine_dir'.
+    The OEDB database can be viewed here:
+    https://openenergy-platform.org/dataedit/view/supply/turbine_library
+    (2019-07-22)
+    Only one turbine configuration is downloaded at a time, if the 
+    search parameters yield an ambigious result, not data is downloaded.
+
```
coroa: Mention in the docs that one can get a table of all turbines at
https://openenergy-platform.org/dataedit/view/supply/turbine_library

atlite/wind.py
coroa: Pretty much all turbines in the [turbine_library](https://openenergy-platform.org/dataedit/view/supply/turbine_library) seem to have multiple hub_heights. Let's warn and take the median of all options for now. I'll try to find out, who is curating them and what the issue behind that is.

coroa: Puh, that comes right out of the manufacturer's reference sheets: for instance, for Enercon: https://www.enercon.de/fileadmin/Redakteur/Medien-Portal/broschueren/pdf/ENERCON_Produkt_de_042017.pdf

atlite/wind.py
coroa: Duh, ok, that is already in the doc string. Sorry for the noise.

coroa:
Maybe replace the start of `download_turbineconf` with something like

```python
def download_turbineconf(turbine=None, store_locally=False, **params):
    if isinstance(turbine, str):
        try:
            params["id"] = int(turbine)
        except ValueError:
            if "_" in turbine:
                params["manufacturer"], params["name"] = turbine.split("_", 2)
            else:
                params["name"] = turbine
    elif isinstance(turbine, dict):
        turbine = turbine.copy()
        turbine.update(params)
        params = turbine

    global _oedb_turbines

    if _oedb_turbines is None:
        [... download ...]

    logger.info("Searching turbine power curve in OEDB database using " +
                ", ".join(f"{k} '{v}'" for (k,v) in params.items()) + ".")
    df = _oedb_turbines[reduce(lambda x, y: x & y,
                               (_oedb_turbines[k] == v for (k,v) in params.items()),
                               pd.Series(True, _oedb_turbines.index))]
```

atlite/wind.py
euronion: Thanks, I knew there was a single line syntax.

atlite/wind.py
euronion: I think multiline suggestions are still not implemented by GitHub (https://github.blog/2018-11-01-suggested-changes-update/).

I rewrote it using a f-string, since we are now supporting only 3.6+

atlite/wind.py
euronion: Good idea!

atlite/wind.py
euronion: I see the appeal, but I am against it limiting ourselves then only to id/name/manu. .
What would you say if we allow for any field from the OEDB and if the provided data is insufficient for a unique retrieve we fail loud?

atlite/wind.py
euronion: I would presonally argue that this would clutter the log too much (think of between 2-~100 turbines which we list here).
And what to list? The full data? Or only ID + Name + Manu.?

atlite/wind.py
euronion: * Printing the hub heights makes sense
* `logging.warn(...)` is deprecated, use `logging.warning(...)` instead (see https://docs.python.org/3/library/logging.html#logging.Logger.warning)

Multiple hub heights are quite common and not that suprising. Depending on the project and location you would want / have to vary them.
I'll take the average but also preserve the original list... .

euronion: + I decided to go with the mean instead of median value.

atlite/wind.py
coroa: Then show:
```python
ds[['id', 'name', 'manufacturer']].head(3)
```

atlite/resource.py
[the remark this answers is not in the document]
euronion: We can raise a `AmbigiouslySpecifiedTurbineException`.

If we continue the process with `None`, then I think the conversion will just fail unexpectedly instead of generating a layout.

I simplified it a bit further, removed 'oedb:' from the string argument in favour of a argument `source='local'|'oedb'`.

euronion:
I prefer to *KISS* here, albeit your solution is elegant.
Especially as I think this flexibility makes the function harder to understand.

atlite/resource.py
```diff
@@ -37,37 +37,39 @@
...
-    if turbine.startswith('oedb:'):
-        source = 'oedb:'
-        turbine = turbine[5:] # Remove 'oedb:' switch
-        turbine = ast.literal_eval(turbine)
-
-        return download_turbineconf(turbine, store_locally=False)
-    elif source is "local":
+    if source.lower() == 'oedb':
+        turbineconf = download_turbineconf(turbine, store_locally=False)
+    elif source == "local":
```
coroa: `source.lower()` should also be used for the check against "local".
I'd propose a
```python
source = source.lower()
```
in front of the if-statements instead

atlite/wind.py
coroa: It's not limiting if it is mainly intended as an additional convenience feature.

I'd allow:
```python
# As convencience all equivalent
download_turbineconf(13)
download_turbineconf({'id':13})
# Maybe as well: download_turbineconf(id=13)

download_turbineconf('E-53/800')
download_turbineconf({'name': 'E-53/800'})
# Maybe as well: download_turbineconf(name='E-53/800')

download_turbineconf('Enercon_E-53/800') # not incidentally exactly the name the turbine would be saved to (except for replacements)
download_turbineconf({'manufacturer': 'Enercon', 'name': 'E-53/800'})
# Maybe as well: download_turbineconf(manufacturer='Enercon', name='E-53/800')
```

I'd gladly support all fields in the second or third manner, but would reserve the simple string/int for `id`, `name` and `manufacturer`/`name`.

atlite/wind.py
```diff
@@ -131,49 +135,48 @@ def download_turbineconf(turbine=None, store_locally=True):
...
+    logger.info("Searching turbine power curve in OEDB database using " +
+                ", ".join(f"{k}='{v}'" for (k,v) in turbine.items()) + ".")
+    # Working copy
+    df = _oedb_turbines
+    if turbine.get('id'):
+        df = df[df.id == int(turbine['id'])]
+    if turbine.get('name'):
+        df = df[df.name == turbine['name']]
+    if turbine.get('manufacturer'):
+        df = df[df.manufacturer == turbine['manufacturer']]
```
coroa: I'd also support other fields like `wind_zone_dibt` and `wind_class_iec`, probably with a `contains` check instead:
```python
for field in ('wind_zone_dibt', 'wind_class_iec'):
    if turbine.get(field):
        df = df[df[field].str.contains(turbine[field])]
```

atlite/resource.py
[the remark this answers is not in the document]
coroa: I still like the `oedb:` prefix together with the possibility to use simple strings for the most usual queries. If you want to keep your `source` argument, you should also provide a way to accept it in `cutout.wind`.

atlite/resource.py
```diff
@@ -29,15 +29,45 @@
...
+
+    if source.lower() == 'oedb':
+        turbineconf = download_turbineconf(turbine, store_locally=False)
+    elif source == "local":
+        res_name = os.path.join(config.windturbine_dir, turbine+".yaml")
+        res_name = construct_filepath(res_name)
+
+        turbineconf = yaml.safe_load(resource_stream(__name__, res_name))
+    else:
+        raise Exception("Not a valid 'source'.")
```
coroa:
```suggestion
        raise ValueError("Not a valid 'source'.")
```

atlite/resource.py
```diff
@@ -29,15 +29,45 @@
...
+    
+    if turbineconf is None:
+        raise Exception("No matching turbine configuration found.")
```
coroa:
```suggestion
        raise ValueError("No matching turbine configuration found.")
```

atlite/resource.py
coroa: https://docs.python.org/3/library/exceptions.html#ValueError

atlite/wind.py
euronion: If you find the curating person, please tell per that I would appreciate a more consistent format :) 
e.g. ideally not `hub_height` entries with partially with a trailing semicolon: "123", "123;" "123; ", "123; 124;"

atlite/wind.py
euronion: Ok, well. The examples do make it clearer, thanks!
I think it is still too much feature and too less simple, but if you insist... .

It should be possible now to use the example arguments you provided above.

I would not implement the selection based on other fields at the moment: The OEDB is not well enough curated at that point and it would be easier to improve the quality of their DB than writing a working parser.

atlite/wind.py
coroa: Sure! :)

atlite/wind.py
coroa: Ok, future problem, then!

atlite/wind.py
```diff
@@ -90,3 +102,158 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+            turbine['id'] = turbine.get('id', int(m[1]))
+            s = "" # Consume the string after success
+            parsed = True
+
+        # 'turbine' is a name or combi of manufacturer + name
+        # Matches e.g. "TurbineName", "Manu1/Manu2[_ ]Turbine/with/number"
+        m = re.match("(?P<manufacturer>[-a-zA-Z0-9\/]*?)?(?:[\s_]*)?(?P<name>[-a-zA-Z0-9\/]+)$", s)
+        if m:
+            turbine['name'] = turbine.get('name', m.group('name'))
+            turbine['manufacturer'] = turbine.get('manufacturer', m.group('manufacturer'))
```
coroa: netither `"name"` nor `"manufacturer"` can be in turbine at this point.

coroa: As a side-note, instead of a pattern like
```
turbine['name'] = turbine.get('name', m.group('name'))
```
better to use the dedicated function
```
turbine.setdefault('name', m.group('name'))
```

atlite/wind.py
```diff
@@ -90,3 +102,158 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
+        # 'turbine' is just a str(-inged) id
+        m = re.match("(^\d+$)", s)
+        if m:
+            turbine['id'] = turbine.get('id', int(m[1]))
+            s = "" # Consume the string after success
+            parsed = True
+
+        # 'turbine' is a name or combi of manufacturer + name
+        # Matches e.g. "TurbineName", "Manu1/Manu2[_ ]Turbine/with/number"
+        m = re.match("(?P<manufacturer>[-a-zA-Z0-9\/]*?)?(?:[\s_]*)?(?P<name>[-a-zA-Z0-9\/]+)$", s)
```
coroa: I'm not a big fan of regular expressions, as they are always hard to read. Allowing to separate manufacturer and name with a space (`\s`) will break if one wants to use turbine names like:
```
download_turbineconf('E-126/4200 EP4')
download_turbineconf({'manufacturer': 'E-126/4200', 'name': 'EP4'})
```

coroa:
Can't help but wonder, whether a slight adaptation of my previous suggestion
```python
    if isinstance(turbine, (int, str)):
        try:
            turbine = {'id': int(turbine)}
        except ValueError:
            turbine = turbine.strip()
            if "_" in turbine:
                manufacturer, name = turbine.split("_", 2)
                turbine = {'manufacturer': manufacturer, 'name': name}
            else:
                turbine = {'name': turbine}
```
is not a lot more readable and concise then your parsing:
https://github.com/FRESNA/atlite/blob/bf72257548ae7e39c6b9b919e42b64a97b5474d9/atlite/wind.py#L140-L170

coroa:
Instead of `name` we should perhaps support the `turbine_type` column, which looks a lot more homogeneous

atlite/wind.py
euronion: I guess I prefer a more gentle way of setting k:v paris in `dict` ;-)

Side note: Right, forgot about `setdefault(...)` .

atlite/wind.py
euronion: The thing is:

If this "convinience feature" allows for `_` to separate and optional manufacturer from a turbine name and turbine name and manufacturer can go ahead an contain special characters like `\s / . `, then the initial argument that "this is the way the file name will look like later" does not hold anymore (because that's what the file name will look like later).

And the `_` has then suddenly be degraded to become "just a separator".
Then we can also just use a real separator which is less commonly used in names, like `| ; ~ # `.
This brings us even further away from the "convinience feature".
If there is a turbine with spaces, slashes, long names, ... , then be all means: Please let us use the dictionary and the user has to input it accordingly.
Then everyone is happy: we don't need to assume much and the users gets what per is asking for (instead of what we make out of the request).

euronion:
> Instead of `name` we should perhaps support the `turbine_type` column, which looks a lot more homogeneous

Agreed, good spot. Also this column does not contain any spaces, so i guess we can keep splitting at `\s` and `_` and go on with our lives. :)

Let's also go with the simple split-and-parse logic.

But man: They really need to add some column descriptiors. Have you found anyone I could kindly ask to do so?

coroa:
Agreed, this has been taken quite a lot more time for refinement than I had hoped for :/. 
If you want to be done with it, i'll finish the integration :), using the snippet from above with `turbine_type` and `manufacturer, name = re.split("[\s_]", s, maxsplit=1)`.

euronion:
Yep, looks like we both found the simpler solution at the same time. :)
See how you like it now.

atlite/wind.py
```diff
@@ -150,19 +150,20 @@ def download_turbineconf(turbine, store_locally=True):
...
         # 'turbine' is a name or combi of manufacturer + name
-        # Matches e.g. "TurbineName", "Manu1/Manu2[_ ]Turbine/with/number"
-        m = re.match("(?P<manufacturer>[-a-zA-Z0-9\/]*?)?(?:[\s_]*)?(?P<name>[-a-zA-Z0-9\/]+)$", s)
-        if m:
-            turbine['name'] = turbine.get('name', m.group('name'))
-            turbine['manufacturer'] = turbine.get('manufacturer', m.group('manufacturer'))
-            s = "" # Consume the string after success
+        # Matches e.g. "TurbineName", "Manu1/Manu2_Turbine/number", "Man. Turb." 
+        # Split on white-spaces and underscore.
+        m = re.split("[\s|_]+",s)
```
coroa: Last nit-pick: Is it intentional to split also on `|`, then you should also reflect this in the comment.
The `maxsplit=1`, i proposed, would additionally make sure you are not falling through the `if .. elif` below.

euronion: Unintentional, I guess coded by habit.
Both suggestions are now included.

coroa:
YAY!

euronion:
"So say we all!"

coroa:
> I wish the turbines in the OEDB would be better documented and described (meta data, consistent format). Right now, we can only hope to access a unique entry using {'id':} or {'manufacturer':, 'name':}.

@euronion: The OEDB turbine library is currently awaiting review: https://github.com/OpenEnergyPlatform/data-preprocessing/issues/26 . Some constructive feedback is probably appreciated!

Issue #27, opened 2019-07-25
[v0.2] Deprecation warnings are ignored by default


coroa:
Uh annoying, they worked wonders in Python 2.7. The default has been changed to ignore them in `3.2` unless you add `-Wd` to your python interpreter or change the filter as you've shown. Hmm .. if I understand the manual correctly it's not intended for libraries to set `captureWarnings` or change the filter. I suppose we should warn in the manual about that and change the examples accordingly.

Do you think `logging.captureWarnings(True)` or `warnings.simplefilter('always', DeprecationWarning)` should be used? I'm tending towards the former.

_Originally posted by @coroa in https://github.com/FRESNA/atlite/pull/23/review_comment/create_

euronion:
There seems to be quite a long story behind it, see e.g. https://lwn.net/Articles/740804/ .

I am in favour of activating them by default either with the `'always'` or `'module'` option.
Users of `atlite` are usually deep in their scripts and we are not talking about encapsulated script behaviour, but relevant API changes.

coroa:
As described in https://github.com/FRESNA/atlite/pull/29#issuecomment-516024659, we should advocate the use of `warnings.simplefilter('default', DeprecationWarning)` to avoid overwhelming users.

euronion:
Sounds good to me.

It is probably sufficient to just include it in the docs and examples, similar to what is already done with `logging` basic configuration.

euronion:
I included sections on this in the documentaion as an example.

Do you think this suffices?
https://atlite.readthedocs.io/en/latest/examples/logfiles_and_messages.html

coroa:
Yes, that looks good.

euronion:
Great. Closing the issue.

PR #29, opened 2019-07-29
convert: Substitute deprecated cutout.meta with cutout.data.

euronion:
*After enabling warnings, I was swarmed with the deprecation warning on `cutout.meta` being deprecated in favour of `cutout.data`.*

Please check if the code intention was correctly transferred.
I did a quick run of the code against my limit scenario and it seems to work.

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -120,8 +120,8 @@ def convert_and_aggregate(cutout, convert_func, windows=None, matrix=None,
 
     if layout is not None:
         if isinstance(layout, xr.DataArray):
-            layout = layout.reindex_like(cutout.meta)
-            layout, _ = xr.broadcast(layout, cutout.meta[['x', 'y']])
+            layout = layout.reindex_like(cutout.data)
+            layout, _ = xr.broadcast(layout, cutout.data[['x', 'y']])
             layout = layout.stack(spatial=('y', 'x')).values
         else:
             assert layout.shape == cutout.shape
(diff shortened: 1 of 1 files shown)


coroa:
_Swarmed_ is never a good sign. The good thing about warnings in respect to the logging module was that every line of code just triggered a single warning and then shut up. Is this behaviour gone?

euronion:
*Swarmed* as in an awkward combination from my side:
* I set the the `simpleFilter('always')`; what you describe is the behaviour of the 'default' option.
This still exists and I would argue should be the option for `atlite` in #27 .
* I tested the wind conversion with my parameter sweeps (some hundreds of calls, each raising the warning).

atlite/convert.py
```diff
@@ -164,7 +164,7 @@ def convert_and_aggregate(cutout, convert_func, windows=None, matrix=None,
     if capacity_factor:
         assert aggregate_func is aggregate_sum, \
             "The arguments `matrix`, `shapes` and `layout` are incompatible with capacity_factor"
-        results /= len(cutout.meta['time'])
+        results /= cutout.data.coords['time'].size
```
coroa:
```suggestion
        results /= cutout.data['time'].size
```

coroa: The `coords` indirection is not useful in this case. There is no clear case for `len` or `.size`, so I don't care.

coroa: Does the suggestion work here? Don't know what the conditions are.

coroa: Bump! Can you apply the modification above?

euronion: Sure, done. Thanks for bumping.
This kind of went under my radar.

PR #39, opened 2019-08-08
shortcut for accessing windturbines and solarpanel

FabianHofmann:
- resource.py add arrowdict for accessing solarpanels and windturbines
- utlis.py add class arrowdict

### DIFF
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -182,11 +182,11 @@ def windturbine_smooth(turbine, params={}):
     return turbine
 
 
-turbines = arrowdict({t[:-5]: t for t in
+turbines = arrowdict({t[:-5]: t[:-5] for t in
                      os.listdir(os.path.join(os.path.dirname(__file__),
                                'resources', 'windturbine'))})
 
-panels = arrowdict({t[:-5]: t for t in
+panels = arrowdict({t[:-5]: t[:-5] for t in
                      os.listdir(os.path.join(os.path.dirname(__file__),
                                'resources', 'solarpanel'))})
 
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -35,7 +35,7 @@ import logging
 logger = logging.getLogger(name=__name__)
 
 from . import config
-from .utils import construct_filepath
+from .utils import construct_filepath, arrowdict
 
 def get_windturbineconfig(turbine):
     """Load the wind 'turbine' configuration.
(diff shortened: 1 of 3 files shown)


coroa:
How is `arrowdict` different from `pypsa.descriptors.Dict`:
https://github.com/PyPSA/PyPSA/blob/9fd2d7413adc5c58825f86d0a5e71616fc12c7f9/pypsa/descriptors.py#L99-L149

 Should we re-use this instead?

coroa:
The dictionaries should be updated by `download_...`

FabianHofmann:
Indeed it is very similar and mainly taken from it. However it is slightly reduced in a way that, when tabbing, _only_ wind turbines (or solarpanels) are proposed for autocompletion, not any dict functions. Also setting attributes is disabled since it's not necessary.

FabianHofmann:
> The dictionaries should be updated by download_...

Makes sense

atlite/resource.py
```diff
@@ -180,3 +180,13 @@ def smooth(velocities, power):
                     "Turbine generates energy at 0 m/s wind speeds".format(p=params))
 
     return turbine
+
+
+turbines = arrowdict({t[:-5]: t[:-5] for t in
+                     os.listdir(os.path.join(os.path.dirname(__file__),
+                               'resources', 'windturbine'))})
```
coroa: The paths used for initialising the dictionaries should take the current configuration into account. So, use the construct_filepath‘ function and initialize it from some function which is called when the config changes. @euronion pbb can add a hint!

atlite/wind.py
```diff
@@ -247,13 +247,15 @@ def download_turbineconf(turbine, store_locally=True):
             turbineconf['HUB_HEIGHT'] = hh[0]
 
     if store_locally is True:
+        from .resource import turbines
```
coroa: I‘d prefer to have a module level import.

coroa:
Thanks looks mostly good. Let’s even import panels and turbines into __init__.py. so that one can use: atlite.turbines....

atlite/resource.py
euronion: For the path to the currently configured windturbine directory, you want to use
```
from . import config
path = construct_filepath(config.windturbine_dir)
```

For updating the function on a config change, you would have to observe the value change (Observer pattern).
I did not design the `config` module with this in mind, I see two ways to make this cleanly

1. Forget about updating the turbines on a config value change.
2. Implement an observer.

Or hack something away, like wrapping this functionality into a function which gets called at strange places.
(Please: don't).

I'd stick with 1. for now and say we make an observer pattern an enhancement request (new issue, low priority).

euronion: also I suggest using `glob.glob(path+"*.yaml")` instead of `os.path(..)` and then `str.replace(".yaml","")` to avoid conflicts with foreign files and non-existing indices.

coroa: And if you are using `glob`, then have a look at the Python 3 [pathlib](https://docs.python.org/3/library/pathlib.html) first!

coroa:
Since @FabianHofmann agrees with the proposed changes in #43. Closing in favour of it.

PR #42, opened 2019-08-12
Continuation of "shortcut for accessing windturbines and solarpanel" aka "Tab-completion for panels and turbines"

coroa:
Continues #39 ...


coroa:
Was against `master` instead of `v0.2`! Closing!

PR #43, opened 2019-08-12
Tab-completion for panels and turbines

coroa:
Supersedes #39 ...


coroa:
@FabianHofmann Had to continue here, since I was missing write access.

Please test and edit how you see fit!

FabianHofmann:
Seems to be fine now! Using globals and an update function makes most sense

coroa:
Ups, sorry, I *always* forget to fix imports.

coroa:
@euronion I touched quite some of your code! Are you fine with the changes?

atlite/resource.py
```diff
@@ -180,3 +184,172 @@ def smooth(velocities, power):
...
+
+        _update_resource_dictionaries()
+        logger.info(f"Turbine configuration downloaded to '{filepath}'.")
+
+
+    return turbineconf
+
+# Global caches
+_oedb_turbines = None
+turbines = arrowdict()
```
euronion:
```suggestion
windturbines = arrowdict()
```
For consistency purposes.

atlite/resource.py
```diff
@@ -180,3 +184,172 @@ def smooth(velocities, power):
...
+panels = arrowdict()
```
euronion:
```suggestion
solarpanels = arrowdict()
```
For consistency purposes.

euronion:
> @euronion I touched quite some of your code! Are you fine with the changes?

@coroa : Looks mostly good, thanks for that easy observer implementation!
The callback now is only activated when calling the `config.update()` method.
This would not activate if I were to directly assign values like `config.windturbine_dir = "/tmp"` .
Can this be fixed easily? Maybe extending on `__setattr__`?

atlite/resource.py
coroa: Food for thought on consistency:
```python
cutout.wind(turbine=atlite.windturbines.Vestas_V112_3MW)
cutout.pv(panel=atlite.solarpanels.CdTe)
```

coroa:
>  Can this be fixed easily? Maybe extending on `__setattr__`?

I don't think `__setattr__` is called on module level (but feel free to try, i am unsure). We could package everything up into an object! :)

An easy solution would be to encourage updating config settings with f.ex.
```python
config.update(windturbine_dir="/tmp")
```
instead.

atlite/resource.py
euronion: `wind(turbine` and `pv(panel` sound nice IMHO.

coroa: Sold!

PR #44, opened 2019-08-13
config: Allow more formats for configuration updates.

euronion:
Following up on #43 .
I will make sure to include this feature into the documentation branch as well.

### DIFF
--- a/atlite/config.py
+++ b/atlite/config.py
@@ -86,7 +86,7 @@ def save(path, overwrite=False):
     with open(path, "w") as config_file:
         yaml.dump(config, config_file, default_flow_style=False)
 
-def update(config_dict = dict(), **kwargs):
+def update(config_dict=None, **kwargs):
     """Update the existing config.
     
     Use a dictionary `config_dict` for updating using a single object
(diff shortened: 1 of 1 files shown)


atlite/config.py
```diff
@@ -86,9 +86,26 @@ def save(path, overwrite=False):
     with open(path, "w") as config_file:
         yaml.dump(config, config_file, default_flow_style=False)
 
-def update(config_dict):
-    """Update the existing config based on the `config_dict` dictionary; resets `config_path`."""
+def update(config_dict = dict(), **kwargs):
```
coroa: Putting pass-by-reference objects into default arguments often leads to unexpected side-effects, since the objects persist.

Here, the `dict()` object will accumulate the kwargs changes over time, ie.:
```python
update(turbine_dir="bla") # will set the turbine_dir config variable to "bla", then
update(panel_dir="blub") # will then set turbine_dir to "bla" and panel_dir to "blub", and so on
```

its better to use a
```python
def update(config_dict=None, **kwargs):
    if config_dict is None:
        config_dict = dict()
```
scheme to reinitialize the dictionaries on each call in such a situation.

Secondly, I would make sure not to update a user-provided dictionary (which will change for the caller, as well):
```python
def update(config_dict=None, **kwargs):
    if config_dict is None:
        config_dict = kwargs
    else:
        config_dict = config_dict.copy()
        config_dict.update(kwargs)
```

euronion: Included. Thanks for pointing it out.

Issue #30, opened 2019-07-30
Parallelised calculations using dask.


euronion:
*Keeping this here as a reminder. Not a pressing issue, but something that has been bothering me for some time.*

I have not yet observed that `atlite` is using more than one processing core for calculations (e.g. wind speed to wind power conversion).
I would assume that this should be easy to implement (as in the backend we are using `dask` and `xarray`, but I haven't been able to identify the culprit.

## Checking for parallelised acitvity
Initially I was only monitoring the CPU util. using `htop`:
There only one core seems to be active.

A better way is to use a `dask` dashboard and a local client:

```
from dask.distributed import Client
import xarray as xr

client = Client()
client
```
I also do not see any dask activity during calculations there.

## Some background info

When checking `cutout.data` this returns an `xarray` Dataset using numpy arrays.
When loading the dataset (`xarray.open_dataset(...) ` ) with the `chunks` option or rechunking the dataset after the cutout was loaded using `cutout.data.chunk(...)`
then shows some dask activity during calculations (but slow, I guess a lot of overhead from spawning and orchestrating the different workers and threads).

## Low priority

With the new version #20 calculations have become significantly faster and caching of datasets obsolete (can now be done implicitly by xarray and dask).
So this is only interesting when doing conversion of large datasets or repeated conversions.

## Current versions
```
> from platform import python_version
> python_version()
3.7.3

> import xarray
> xarray.__version__
'0.12.3'

> import dask
> dask.__version__
'2.1.0'
```

coroa:
The _"culprit"_ is two-fold:
- `windows !== False` (including the default), will construct an iterator DataArrays backed by numpy arrays (dask arrays are explicitly loaded into memory), unless the `@requires_windowed` decorator has `allow_dask=True`.
https://github.com/FRESNA/atlite/blob/25727ceccfddffa9edd39e0e413a0d6144bf00d4/atlite/data.py#L121-L124
- This opt-in mechanism is to help with the transition from pre-dask code. For instance `heat_demand` would not work with dask, as it modifies the `values` attribute (which creates a (read-only??) copy when running on a dask-backed array)
https://github.com/FRESNA/atlite/blob/25727ceccfddffa9edd39e0e413a0d6144bf00d4/atlite/convert.py#L232
instead  this would have to be rewritten as `heat_demand.clip(lower=0.)` instead. The `pv` module has several of these inplace patterns, since they generally run faster than creating copies with `clip`, at the draw back of excluding dask.

[...]

PR #48, opened 2019-09-19
Use dask for PV time-series

coroa:
`convert_pv` is the last hurdle to a full dask mode in atlite. Refer also to #30 .

This PR replaces all the inplace transformations with clip and where so they work in principle on dask.

Unfortunately dask fails to properly optimize the task graph (into something which should be linear in time), but instead thrashes memory until it dies even for more small cutouts.

I wasn't able today to find any working solution (not even in synchronuous mode it works sensibly).

For future reference, its very helpful to visualize and profile dask:
```python
import dask
from dask.diagnostics import Profiler, ResourceProfiler, CacheProfiler, visualize
from dask.callbacks import Callback
class PrintKeys(Callback):
    def _pretask(self, key, dask, state):
        """Print the key of every task as it's started"""
        print("Computing: {0}!".format(repr(key)))

panel = atlite.resource.get_solarpanelconfig("CSi")
orientation = atlite.pv.orientation.get_orientation({"slope": 30, "azimuth": 180})
solar = atlite.convert.convert_pv(cutout.data, panel, orientation)

store_task = solar.to_netcdf('test.nc', compute=False)
store_task_opt, = dask.optimize(store_task)
store_task_opt.visualize()
# or: store_task_opt.visualize(filename="graph.dot")
# and then look at it using xdot

with dask.config.set(num_workers=2), \
     Profiler() as profiler, \
     ResourceProfiler(0.5) as mem, \
     CacheProfiler() as cache, \
     PrintKeys():
        dask.compute(store_task)

# shows bokeh plots for the dask computation
visualize([profiler, mem, cache])
```

You might also want to look at intermediate values, starting from:
```python
from atlite.pv.solar_position import SolarPosition
from atlite.pv.irradiation import TiltedIrradiation
from atlite.pv.solar_panel_model import SolarPanelModel
from atlite.pv.orientation import get_orientation, SurfaceOrientation
ds = cutout.data.astype(np.float32)
solar_position = SolarPosition(ds)
surface_orientation = SurfaceOrientation(ds, solar_position, orientation)

irradiation = TiltedIrradiation(ds, solar_position, surface_orientation, trigon_model="simple",
                                clearsky_model="simple")

solar_panel = SolarPanelModel(ds, irradiation, panel)
```

### DIFF
--- a/atlite/pv/irradiation.py
+++ b/atlite/pv/irradiation.py
@@ -118,14 +118,14 @@ def _albedo(ds, influx):
 
 def TiltedGroundIrrad(ds, solar_position, surface_orientation, influx):
     surface_slope = surface_orientation['slope']
-    ground_t = influx * _albedo(ds, influx) * (1.0 - np.cos(surface_slope)) / 2.0
+    ground_t = influx * _albedo(ds, influx) * (1 - np.cos(surface_slope)) / 2
     return ground_t.rename('ground tilted')
 
-def TiltedIrradiation(ds, solar_position, surface_orientation, trigon_model, clearsky_model, altitude_threshold=1.):
+def TiltedIrradiation(ds, solar_position, surface_orientation, trigon_model, clearsky_model, altitude_threshold=1., dtype=np.float32):
 
     influx_toa = solar_position['atmospheric insolation']
     def clip(influx, influx_max):
-        return influx.clip(min=0., max=influx_max.transpose(*influx.dims))
+        return influx.clip(min=0, max=influx_max.transpose(*influx.dims))
 
     if 'influx' in ds:
         influx = clip(ds['influx'], influx_toa)
@@ -143,10 +143,10 @@ def TiltedIrradiation(ds, solar_position, surface_orientation, trigon_model, cle
 
         influx = direct + diffuse
         direct_t = k * direct
-        diffuse_t = ((1. + cos_surface_slope) / 2. * diffuse +
-                    _albedo(ds, influx) * influx * ((1. - cos_surface_slope) / 2.))
+        diffuse_t = ((dtype(1) + cos_surface_slope) / dtype(2) * diffuse +
+                     _albedo(ds, influx) * influx * ((dtype(1) - cos_surface_slope) / dtype(2)))
 
-        total_t = direct_t.fillna(0.) + diffuse_t.fillna(0.)
+        total_t = direct_t.fillna(0) + diffuse_t.fillna(0)
     else:
         diffuse_t = TiltedDiffuseIrrad(ds, solar_position, surface_orientation, direct, diffuse)
         direct_t = TiltedDirectIrrad(solar_position, surface_orientation, direct)
--- a/atlite/pv/irradiation.py
+++ b/atlite/pv/irradiation.py
@@ -91,7 +91,7 @@ def TiltedDiffuseIrrad(ds, solar_position, surface_orientation, direct, diffuse)
             logger.warn('diffuse_t exhibits negative values above altitude threshold.')
 
     with np.errstate(invalid='ignore'):
-        diffuse_t.values[np.isnan(diffuse_t.values) | (diffuse_t.values < 0.)] = 0.
+        diffuse_t.clip(min=0.).fillna(0.)
 
     return diffuse_t.rename('diffuse tilted')
 
@@ -158,7 +158,8 @@ def TiltedIrradiation(ds, solar_position, surface_orientation, trigon_model, cle
     # values, leading to big overall errors from the 1/sinaltitude factor.
     # => Suppress irradiation below solar altitudes of 1 deg.
 
-    cap_alt = solar_position['altitude'] < np.deg2rad(altitude_threshold)
-    total_t.values[(cap_alt | (direct+diffuse <= 0.01)).transpose(*total_t.dims).values] = 0.
+    cap_alt = solar_position['altitude'] > np.deg2rad(altitude_threshold)
+    total_t = total_t.where(cap_alt & (direct + diffuse > 0.01), dtype(0))
+    # total_t.values[(cap_alt | (direct+diffuse <= 0.01)).transpose(*total_t.dims).values] = 0.
 
     return total_t
--- a/atlite/pv/irradiation.py
+++ b/atlite/pv/irradiation.py
@@ -91,7 +91,7 @@ def TiltedDiffuseIrrad(ds, solar_position, surface_orientation, direct, diffuse)
             logger.warn('diffuse_t exhibits negative values above altitude threshold.')
 
     with np.errstate(invalid='ignore'):
-        diffuse_t.clip(min=0.).fillna(0.)
+        diffuse_t = diffuse_t.clip(min=0.).fillna(0.)
 
     return diffuse_t.rename('diffuse tilted')
 
--- a/atlite/pv/orientation.py
+++ b/atlite/pv/orientation.py
@@ -90,7 +90,7 @@ def SurfaceOrientation(ds, solar_position, orientation):
     # fixup incidence angle: if the panel is badly oriented and the sun shines
     # on the back of the panel (incidence angle > 90degree), the irradiation
     # would be negative instead of 0; this is prevented here.
-    cosincidence.values[cosincidence.values < 0.] = 0.
+    cosincidence = cosincidence.clip(min=0.)
 
     return xr.Dataset({'cosincidence': cosincidence,
                        'slope': surface_slope,
--- a/atlite/pv/solar_panel_model.py
+++ b/atlite/pv/solar_panel_model.py
@@ -34,8 +34,7 @@ def _power_huld(irradiance, t_amb, pc):
                      pc['k_5'] * (np.log(G_)) ** 2) +
                pc['k_6'] * (T_ ** 2))
 
-        eff = eff.fillna(0.)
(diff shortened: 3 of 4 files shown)


euronion:
I tested variants of this version with `cutout.wind(...)`, `cutout.pv(...)` and `cutout.heat_demand(...)`.
All seem to work nicely and produce the same results (within the tested scope) as the old modules.

## Limitations / Restrictions

What is still missing is an appropriate place to introduce (automatic) chunking of the cutout, otherwise dask is not feeling responsible. For now I used 
```
cutout.data = cutout.data.chunk({d:'auto' for d in cutout.data.dims.keys()})
```
after loading the cutout.

Also the `index` parameter here
https://github.com/PyPSA/atlite/blob/88f2e2b0a2d87dd60b0ff3330abad040f3f52937/atlite/convert.py#L51-L55

is stricter for using the dask mode:
Without the dask backend `index` can be provided as a `list`.
With the dask backend this fails with a rather difficult to track down error caused in the aggregation step. If one follows the documentation and provides a `pandas.Index` everything is fine though.
Not sure if this requires some changes to e.g. conserve the current (undocumented) behaviour for a `list`.

## Workers exceeding allowed memory pool

I can confirm the problem with the `pv` conversion requiring significant amounts of memory. I do not get any errors when using the default scheduler and settings (as sufficient memory is available), but the spike is there and it barely fits.
I tracked it down to be linked somehow to the final computation where all the different data is combined. It is also linked to the chunking used: I get the same large memory requirements 6 times, once for each chunk (time is auto-chunked into 6 equally large chunks of 1,464 timestemps).
I'll try to track it down further ... .

## Obsolete code pieces?
If we figure this out, am I correct to assume that we can remove the whole `requires_windowed` decoator? It would make the code significantly simpler.

Also we can probably make the aggregation here
https://github.com/PyPSA/atlite/blob/88f2e2b0a2d87dd60b0ff3330abad040f3f52937/atlite/aggregate.py#L29
simpler, as we do not need any distinction between backends?

atlite/pv/irradiation.py
```diff
@@ -91,7 +91,7 @@ def TiltedDiffuseIrrad(ds, solar_position, surface_orientation, direct, diffuse)
             logger.warn('diffuse_t exhibits negative values above altitude threshold.')
 
     with np.errstate(invalid='ignore'):
-        diffuse_t.values[np.isnan(diffuse_t.values) | (diffuse_t.values < 0.)] = 0.
+        diffuse_t.clip(min=0.).fillna(0.)
```
coroa:
```suggestion
        diffuse_t = diffuse_t.clip(min=0.).fillna(0.)
```

coroa:
Did you test `cutout.pv(..., windows=False)`? Otherwise there is not much dask involved, for now.

In principle the decorator could go then. The code simplification is relatively local, though. Only parts of `data.py` would not be needed anymore.

But I would want to be sure, we are not introducing major performance degradations by relying on dask only.

The `aggregate_matrix` branch introduced for dask, should actually also work for non-dask arrays. Again I am not sure, how it would behave performance wise, since the matrix product is applied to a transposed array (probably minor changes only, I suspect, but haven't tried).

An `index = pd.Index(index)` line in `aggregate_matrix` would not hurt to bring back the old behaviour when `isinstance(index, list)`.

euronion:
Yes, I checked with `windows=False`, `dask` is definitley taking care of the calculations:
Issuing the `cutout.pv(...)` command returns nearly immediately with a `dask.array` wrapped by a`xarray.DataArray` :
```
> pv = cutout.pv("CSi",{"slope": 30, "azimuth": 180}, windows=False)
> pv
<xarray.DataArray (y: 133, x: 157)>
dask.array<shape=(133, 157), dtype=float64, chunksize=(133, 157)>
Coordinates:
  * y        (y) float64 44.0 43.75 43.5 43.25 43.0 ... 11.75 11.5 11.25 11.0
    lat      (y) float64 dask.array<shape=(133,), chunksize=(133,)>
  * x        (x) float64 25.0 25.25 25.5 25.75 26.0 ... 63.25 63.5 63.75 64.0
    lon      (x) float64 dask.array<shape=(157,), chunksize=(157,)>
```
as it should be.

Performance wise I feel like there is a bit overhead (not much, I'll report numbers later). The simplifications gained by understanding the code (when reading because e.g. of bugtracking) would I think be more valuable than minor performance loses.

A actually prefer the `pd.Index` compared to a list, as it gives more sense to the return values. The error message for a `list` however is not really helpful. I'd prefer a deprecation warning or a `TypeError`.

euronion:
## Addendum - timings

Timing using `%%time` on the following code

```
pv = cutout.pv("CSi",{"slope": 30, "azimuth": 180})
pv.compute()
print(pv)
```

cutout dimensions (ERA5):
`'x': 157, 'y': 133, 'time': 8784`

For the codebase, *GitHub* refers to the `daskpv` branch, the *Dask-specific code* has the first code block enabled by `if True:`, the *Non-Dask code*  has the second code block enabled by `if False:` in `aggregate.py`:

| Codebase | Chunked | `windows` | `dask` | `num_workers` | Timing |
|----------- | ---------- | ----------- | ------- | ---------------- | -------- |
| GitHub | Yes | `True` | `enabled` | auto | 12.3s |
| GitHub | Yes | `True` | `enabled` | 1 | 31.4s |
| GitHub | Yes | `None` (default) | `off` | - | 41.9s |
| GitHub | No | `None` (default) | `off` | - | 41.5s |
| Dask-specific code | Yes | `None` (default) | `off` | - | 41.8s | 
| Non-Dask code | Yes | `True` | `off` | - | 41.3s |
| Non-Dask code | Yes | `False` | `on` | auto |12.5s |

I do not see any significant changes.
The only question still open is, why there is a different approach for dask arrays / other backends it in the first place?

euronion:
side note:
https://github.com/PyPSA/atlite/blob/8efc4c8f16e1244e2af3dd04137a2f09991c2a01/atlite/pv/irradiation.py#L113

should be
```
albedo = albedo.where(albedo <= 1., 1.)
```

euronion:
## Update on memory usage
I think this is related to `dask` poorly handling intermediate values and keeping too much in memory. One solution to the problem is described in the [dask documentation](https://docs.dask.org/en/latest/array-creation.html#intermediate-storage) and as point 2 in the [xarray documentation](http://xarray.pydata.org/en/stable/dask.html#optimization-tips). Also presumable also this [long open issue](https://github.com/dask/dask/issues/874).

I was able to trim it down (~50% for my case from ca. 60GB to 32GB) using 5-6 `dask.persist()` statements in intermediary steps (placement rule: following my gut feeling).
These trigger the computation of intermediary results and seems to also allow `dask` to better optimise the task graph.



## Task graphs
(External links because of large svg files - sorry)

* before: https://svgshare.com/i/FSa.svg
* after: https://svgshare.com/i/FSE.svg

Caveat:
It did degrade performance a bit in comparison to the case without `.persist()` statements. The speedup on my rig was still ~2x-3x compared to the non-dask situation.

Do you think that's a reasonable solution? I'd like to finish this up, already literally a week on this without much success. It's a pain in the seat muscles to pinpoint and solve.

euronion:
## Update - Alternative

I didn't really like the idea of adding `persist()` into the code, so I followed the recommendations and added some temporary flushing to disk. More precisley I changed

https://github.com/PyPSA/atlite/blob/b227ab23383c1160b870b78ac0a5d92ee864fca2/atlite/convert.py#L414-L421

to

```
414 imoprt os, tempfile
415 def convert_pv(ds, panel, orientation, trigon_model='simple', clearsky_model='simple'):
416
417     fh, fn = tempfile.mkstemp()
418     os.close(fh)
419
420     solar_position = SolarPosition(ds)
421     surface_orientation = SurfaceOrientation(ds, solar_position, orientation)
422
423     d_surface_orientation = surface_orientation.to_netcdf(fn, compute=False)
424     surface_orientation = xr.open_dataset(fn)
425
426     irradiation = TiltedIrradiation(ds, solar_position, surface_orientation,
427                                     trigon_model=trigon_model,
428                                     clearsky_model=clearsky_model)
429     solar_panel = SolarPanelModel(ds, irradiation, panel)
430
431     surface_orientation.close()
432     os.remove(fn)
433
434     return solar_panel
```

This yields similar gains in memory reduction (down to 20-25 GB from ~60 GB) in my case, but is noticeably slower than the `persist()` approach.

euronion:
## Update - Better alternative

I was just trying out another alternative, which is now my favourite one:
https://github.com/euronion/atlite/blob/14633a7a86d130fc2d864dcd8de719a522aa400a/atlite/convert.py#L414-L427

See [this branch](https://github.com/euronion/atlite/tree/daskpv-mod001/atlite).

Results:
* The first `cutout.pv(...)` initially takes similarly long as without the whole dask mode [because loading from disk is done repeadetley]
* The second `cutout.pv(...)` call is significantly faster now
* `pv = cutout.pv(...)` calls still return a delayed action, as the `convert_pv(..)` action is computed but not the `aggregation_...(...)` method.
* Gains from using dask are still present (albeit a bit dampened).

I am now down to 11 GB of memory usage (reduction by a factor of 5 to 6).

Could you confirm this to work for you?

fneum:

euronion:
Yes.

coroa:
Nope, out of scope

FabianHofmann:
Closing this in favour of code in #20

Issue #45, opened 2019-08-13
Implement configuration as object instead of module.


euronion:
We do have (with the new version #20 ) a new configuration system in the pipeline.

It does however have a few downsides which became apparent during further developments, see e.g. the discussion here, where an observer can not trigger on changes of the module variables.

>> I don't think `__setattr__` is called on module level (but feel free to try, i am unsure).
>> We could package everything up into an object! :)
>
> Thought for the future:
> I worked on it in https://github.com/euronion/atlite/blob/f1e039abf9c5295ac549d0c39152beff7697c418/atlite/config.py
> and had a few things I could not figure out and dropped it.
> (like not calling the config as `atlite.config.config` and making sure the object is shared > among all modules when a new config-object is used to overwrite the existing one - in short:
> The current solution is more robust).

_Originally posted by @euronion in https://github.com/PyPSA/atlite/pull/43#issuecomment-520755345_

It would be nice to implement the config as a true object.

coroa:
A serializable object would probably also help in distributed settings. In the current scheme, when using `from dask.distributed import Client; Client()`, the workers only see the default configuration, since they live in separate python processes.

euronion:
We're not using a config object or configuration file anymore but instead opted for function arguments.

PR #50, opened 2019-10-30
Update config system to use a object rather than a module.

euronion:
> Implementing https://github.com/PyPSA/atlite/issues/45

The config system now uses a config object, located in the module "config".
The parent class is atlite.config.Config, access to the configuration is now achieved via atlite.config.config .
Getters and setters are implemented, the callback function for update hooks is mostly working, except for when the object atlite.config.config is overwritten directly with a new Config object.

Since we can now use getters and setters, manual calls for resolving the (relative) file path are obsolete and now automatically handled by the `config` object.

I could not figure out two problems related to direct overwrites of the `atlite.config.config` object, i.e. by
```
atlite.config.config = Config()
```
1. How to call the `_UPDATE_HOOKS` (listener pattern on module variable)
2. How to use the config in other modules as
```
from .config import config
...
config.windturbine_dir
```
without comprimising the integrity in the case of a direct assignment above.
The current version works in this case, i.e. the complexity
```
from . import config
....
config.config.windturbine_dir
```
is required for covering the case that the `config` object changes.


fneum:

euronion:
It was, not sure if the changes by Jonas made it obsolete - I'll have to check it during review.

coroa:
Obsolete

coroa:
No config system, instead.

PR #51, opened 2019-11-10
Config object counter-proposal

coroa:
Different take on the config object.

Load from config:
```python
config = atlite.Config.from_file("blabla.yaml")
config = atlite.Config.from_dict({...})
cutout = atlite.Cutout(..., config=config)

# as a short-cut it's also possible to
cutout = atlite.Cutout(..., config="blabla.yaml")
```

Updates:
```python
cutout.config.windturbine_dir = "..."
cutout.config.gebco_path = "..."
# or
cutout.config.update(gebco_path=...)
```
Windturbines can be completed with
```python
cutout.resource.windturbines.Vestas...
```

```python
atlite.windturbines
```
is still supported, but only shows windturbines in the default config path.


atlite/config.py
```diff
@@ -1,146 +1,199 @@
...
+        Use a dictionary `config_dict` for updating using a single object
+        or provide assignment expressions to the config variables you
+        want to change.
+        Using this method ensures that all internal dependencies relying
+        on path information are also also correctly updated.
+
+        Parameters
+        ----------
+        config_dict : dict
+            (Default: dict()). Dictionary with pairs of key value pairs:
```
euronion:
```suggestion
            Dictionary with pairs of key value pairs:
```

atlite/config.py
```diff
@@ -1,146 +1,199 @@
...
+            <config var. name>:<new value>.
+        **kwargs
+            Any existing or new config variable and its new value to store
+            in the atlite.config.
```
euronion:
```suggestion
```

atlite/config.py
```diff
@@ -1,146 +1,199 @@
...
+
+        for key, val in updates.items():
+            setattr(self, f"_{key}", val)
+
+        for func in self._UPDATE_HOOKS:
+            func(self)
+
+    def register_update_hook(self, func):
+        self._UPDATE_HOOKS.append(func)
+        func(self)
```
euronion:
```suggestion
```
Registering a hook shouldn't implicilty trigger the function call.

atlite/config.py
```diff
@@ -1,146 +1,199 @@
...
+    def read(self, path):
+        """Read and set the configuration based on the file in 'path'."""
+
+        path = Path(path)
+        if not path.is_file():
+            raise TypeError(f"Invalid configuration file path {path}.")
+
+        with open(path, "r") as config_file:
+            config_dict = yaml.safe_load(config_file)
+            config_dict.setdefault("config_path", path.parent)
```
euronion:
```suggestion
            config_dict.setdefault("config_path", path)
```

atlite/config.py
```diff
@@ -1,146 +1,199 @@
...
+            (Default: None)
+        """
+
+        for var in self._VARIABLES:
+            setattr(self, f"_{var}", None)
+
+        # Prefer config explicitly provided as a dict
+        if isinstance(config_dict, dict):
+            logger.info(f"Loading config from dict.")
+            config_dict.setdefault("config_path", ".")
```
euronion:
```suggestion
            config_dict.setdefault("config_path", _FILE_SEARCH_NAME)
```

For consistency, let's refer with `config_path` to the path of the configuration file, not its parent directory.

atlite/convert.py
```diff
@@ -401,7 +398,7 @@ def wind(cutout, turbine, smooth=False, **params):
     """
 
     if isinstance(turbine, str):
-        turbine = get_windturbineconfig(turbine)
+        turbine = self.resource.get_windturbineconfig(turbine)
```
euronion: This will be a not backwards compatible change. Are we okay with that?

atlite/cutout.py
```diff
@@ -60,8 +61,8 @@ def __init__(self, name=None, data=None, cutout_dir=None, **cutoutparams):
         if dirname:
             cutout_dir = dirname
         elif cutout_dir is None:
-            if config.cutout_dir:
-                cutout_dir = utils.construct_filepath(config.cutout_dir)
+            if config.config.cutout_dir:
```
euronion:
```suggestion
            if self.config.cutout_dir:
```

atlite/cutout.py
```diff
@@ -60,8 +61,8 @@ def __init__(self, name=None, data=None, cutout_dir=None, **cutoutparams):
...
+                cutout_dir = config.config.cutout_dir
```
euronion:
```suggestion
                cutout_dir = self.config.cutout_dir
```

euronion:
```suggestion
                cutout_dir = self.config.cutout_dir
```

Here and in the line above it should be `self.config.cutout_dir`.

However:
Making the config part of the cutout should make the attribute `cutout.cutout_dir` obsolete.
We can keep it in the descriptor, but we should at this point - if not set - `self.config.cutout_dir = cutout_dir`
and substitute subsequent calls accordingly.

coroa: Good idea,

I cleaned up all uses of `cutout_dir` and removed the variable from the constructor's call signature. Instead I introduced a `Cutout.cutout_path` which points to the netcdf file or old-style directory the cutout is based on.

coroa:
I addressed most of your suggestions, i think.

The datasets era5 and sarah should work with the new changes; ncep and cordex need to be rewritten, anyway, i'm only keeping them as a reminder.

Saving the config with the cutout is not easily practical. Since the config is mostly concerned with paths for finding and generating cutouts, we have a hen egg problem. Together with the path of the config file defines default search path rule it might get a bit complicated.

atlite/resource.py
```diff
@@ -36,90 +36,252 @@
...
+
+        def as_turbineconfig(turbine):
+            if isinstance(turbine, str):
+                if turbine.startswith('oedb:'):
+                    turbine = turbine[len("oedb:"):]
+                    return self.download_windturbineconfig(turbine, store_locally=False)
+                else:
+                    return read_windturbineconfig(turbine)
+            elif isinstance(turbine, dict):
+                if turbine.get('source') is None:
```
euronion: The source keyword now no longer allows to specify a config directly as a dict.

I.e. this:

`cutout.get_windturbineconfig({"V":[1,2,3], "P":[4,5,6], "HUB_HEIGHT":42})`

is no longer possible, as per default "oedb" is assumed or I could specify "local".
I'd like to either have a dedicated keyword (e.g. 'manual') or have let default (no 'source') be directly as a dict.

atlite/cutout.py
```diff
@@ -35,38 +35,68 @@
...
-    dataset_module = None
+def ensure_path(name, has_data, config):
+    """
+    Translate `name` argument to the full path of a cutout netcdf file or
+    directory of an old-style cutout
 
-    def __init__(self, name=None, data=None, cutout_dir=None, **cutoutparams):
+    In order, we look for:
+
+    - File with nc suffix in current working directory
```
euronion: Why would we prioritise the current working directory, even if a `cutout_dir` is specified?
Shouldn't it be the other way round?

atlite/cutout.py
```diff
@@ -35,38 +35,68 @@
...
 
 from .convert import (convert_and_aggregate, heat_demand, hydro, temperature,
                       wind, pv, runoff, solar_thermal, soil_temperature)
+from .resource import Resources
 from .gis import GridCells
 from .data import requires_coords, requires_windowed, cutout_prepare
 
-class Cutout(object):
-    dataset_module = None
+def ensure_path(name, has_data, config):
```
euronion: Could we document, what the meaning of `has_data` as a switch is?

euronion:
> Saving the config with the cutout is not easily practical. Since the config is mostly concerned with paths for finding and generating cutouts, we have a hen egg problem. Together with the path of the config file defines default search path rule it might get a bit complicated.

I see, then let's leave it be.

We also need to update the documentation to reflect the new configuration and should there mention how the saving works / when it needs to be done.

atlite/resource.py
coroa: How would this work in the original code:
https://github.com/PyPSA/atlite/blob/045cda6bac0a70df2932ade373607d5b5953d83b/atlite/resource.py#L63-L83

atlite/utils.py
```diff
@@ -55,10 +55,12 @@ def make_optional_progressbar(show, prefix, max_value=None):
...
-def migrate_from_cutout_directory(old_cutout_dir, name, cutout_fn, cutoutparams):
-    old_cutout_dir = Path(old_cutout_dir)
-    with xr.open_dataset(old_cutout_dir / "meta.nc") as meta:
-        newname = old_cutout_dir.parent / f"{name}.nc"
+def migrate_from_cutout_directory(cutout_path, cutoutparams, config):
+    cutout_dir = config.cutout_dir
+    if cutout_dir is None:
+        cutout_dir = cutout_path.parent
+    new_cutout_path = cutout_dir / f"{cutout_path.stem}.nc"
+    with xr.open_dataset(cutout_path / "meta.nc") as meta:
```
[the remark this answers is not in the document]
coroa: No cutout_path points to the directory of the previous cutout. `cutout_dir` is the directory where to save the migrated cutout.

atlite/utils.py
```diff
@@ -78,7 +80,7 @@ def migrate_from_cutout_directory(old_cutout_dir, name, cutout_fn, cutoutparams)
         """))
 
         try:
-            data = xr.open_mfdataset(str(old_cutout_dir / "[12]*.nc"), combine="by_coords")
+            data = xr.open_mfdataset(cutout_path / "[12]*.nc", combine="by_coords")
```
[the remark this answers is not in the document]
coroa: No, as above

atlite/resource.py
euronion: You're right. It wouldn't have worked back then. No need to change it then either.

fneum:

euronion:
It was, not sure if the changes by Jonas made it obsolete - I'll have to check it during review.

coroa:
Nope, no conf system instead.

euronion:
Different approach used in `v0.2` . PR obsolete.

PR #63, opened 2020-04-29
convert: Fix heat demand hourshift for xarray 0.15.1

nworbmot:
From xarray version 0.15.1, .values cannot be assigned. You should
use the .assign_coords() method instead.

See release notes for xarray 0.15.1 "breaking changes":

http://xarray.pydata.org/en/stable/whats-new.html#v0-15-1-23-mar-2020


nworbmot:
This change is also backwards-compatible with older versions of xarray (at least with version 0.12).

PR #69, opened 2020-05-10
use shapely's strtree instead of rtree

FabianHofmann:
Shapely provides its own 'rtree' for filtering out shapes which cannot intersect with each other. 

This change includes the [shapely.ops.strtree functionality](https://shapely.readthedocs.io/en/latest/manual.html#strtree.STRtree). Therefore the backup code in `compute_indicatormatrix` is obsolete. 

The speed gain is about 90%

### DIFF
--- a/atlite/gis.py
+++ b/atlite/gis.py
@@ -134,41 +134,6 @@ def compute_indicatormatrix(orig, dest, orig_proj='latlong', dest_proj='latlong'
 
     return indicator
 
-class GridCells:
-    def __init__(self, grid_cells, sindex, projection):
-        self.grid_cells = grid_cells
-        self.sindex = sindex
-        self.projection = projection
-
-    @classmethod
-    def from_cutout(cls, cutout):
-        coords = cutout.grid_coordinates()
-        span = (coords[cutout.shape[1]+1] - coords[0]) / 2
-        grid_cells = [box(*c) for c in np.hstack((coords - span, coords + span))]
-        sindex = Index((j, o.bounds, None) for j, o in enumerate(grid_cells))
-
-        return cls(grid_cells, sindex, cutout.projection)
-
-    @staticmethod
-    def from_file(filename):
-        pass
-
-    def to_file(self, filename):
-        # Turns out the spatial index gets distorted by pickling, so we turn this
-        # into a no-op, until we figure out a work-around
-        pass
-
-    def indicatormatrix(self, shapes, shapes_projection='latlong'):
-        shapes = reproject_shapes(shapes, shapes_projection, self.projection)
-        indicator = sp.sparse.lil_matrix((len(shapes), len(self.grid_cells)), dtype=np.float)
-
-        for i, s in enumerate(shapes):
-            for j in self.sindex.intersection(s.bounds):
-                o = self.grid_cells[j]
-                area = s.intersection(o).area
-                indicator[i,j] = area/o.area
-
-        return indicator
 
 def maybe_swap_spatial_dims(ds, namex='x', namey='y'):
     swaps = {}
--- a/atlite/gis.py
+++ b/atlite/gis.py
@@ -27,6 +27,7 @@ import rasterio as rio
 import rasterio.warp
 from rasterio.warp import Resampling
 from rtree.index import Index
+from shapely.strtree import STRtree
 
 import logging
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


coroa:
Great, I didn't know about this! Awesome that you see that much of a speed improvement about this last quite annoying bit of extra time. (If you have time to spare maybe propose a PR to geopandas to add this to their https://github.com/geopandas/geopandas/blob/master/geopandas/sindex.py, although now I see they included PyGEOS as an alternative).

Before you rip out GridCells #70  . Can you please time for a realistically sized cutout the time it takes to generate the boxes (ie a call to grid_cells) plus the generation of the indicator matrix. Because i think the memory allocation and creation of these `shapely.geos.box` objects also takes a signifcant amount of time (which is also cached in the GridCells object).

FabianHofmann:
Sounds fair! Here a small comparison for a cutout with 2068 grid cells


### Create grid cells (without creation of sindex)
```python 
%timeit atlite.gis.GridCells.from_cutout(cutout)
341 ms ± 20.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
``` 


### Create grid cells + indicator matrix without cached grid cells 
```python 
%%time
cells = cutout.grid_cells
dest = shapes.geometry
indicator_matrix = atlite.gis.compute_indicatormatrix(cells, dest)
CPU times: user 3.52 s, sys: 14.2 ms, total: 3.53 s
Wall time: 3.53 s
```
### Create grid cells + indicator matrix with cached grid cells 
```python
%%time
cells = cutout.grid_cells
dest = shapes.geometry
indicator_matrix = atlite.gis.compute_indicatormatrix(cells, dest)
CPU times: user 3.3 s, sys: 1.97 ms, total: 3.3 s
Wall time: 3.29 s
```

The difference with and without cached grid cells is  ~0.25 seconds. I think its fine to cach them (I don't have a strong opinion here). 
However the current caching is mainly out of the `GridCells` class. One can just replace the current underlying `_grid_cells` function by 
```python 
    _grid_cells_cache = None
    @property
    def grid_cells(self):
        if self._grid_cells_cache is not None:
            return self._grid_cells_cache

        coords = self.grid_coordinates()
        span = (coords[self.shape[1]+1] - coords[0]) / 2
        grid_cells = [box(*c) for c in np.hstack((coords - span, coords + span))]

        # Cache
        self._grid_cells_cache = grid_cells

        return grid_cells
``` 
Which would probably take up your point?

coroa:
> Which would probably take up your point?
Yes, fine.

Why is the time `3.3s` in total. I thought you said, index construction went down to a couple of `ms`?

FabianHofmann:
alright, just for notice, only the `sindex` creation is replaced by the performant `srtrtee`. Not the whole indicator matrix creation (sorry for the disclaimer :-( ) 

I will push further changes here in order to remove the `GridCells` class.

FabianHofmann:
@euronion 
For chaching properties, [this](https://docs.python.org/3/library/functools.html#functools.cached_property) decorator would be perfect. However it's only introduced in python 3.8. The lru_cache is for functions only, not properties. So we stick to the manual cache. 

The new behaviour is tested in `test_indicator_matrix` in the test script `test_cutout_creation.py`.

coroa:
> For chaching properties, [this](https://docs.python.org/3/library/functools.html#functools.cached_property) decorator would be perfect

Or `CachedAttribute` which has been available in vresutils since 2015. https://github.com/FRESNA/vresutils/blob/5d5fbb5d19454042c5bc1c379b4634d1182cfd6a/vresutils/decorators.py#L187-L208

atlite/gis.py
```diff
@@ -121,67 +122,18 @@ def compute_indicatormatrix(orig, dest, orig_proj='latlong', dest_proj='latlong'
...
-        idx = Index()
-        for j, o in enumerate(orig):
-            idx.insert(j, o.bounds)
-
-        for i, d in enumerate(dest):
-            for j in idx.intersection(d.bounds):
-                o = orig[j]
+    for i, d in enumerate(dest):
+        for o in tree.query(d):
+            if o.intersects(d):
```
coroa, line 130: Did you test whether the inclusion of this `if` branch speeds things up?

Because with rtree, it was actually slower to test the intersection first and the result does not change since if they dont intersect the intersection area is 0 and `lil_matrix` suppresses zero values.

FabianHofmann: yes it speeds up, about 25 %. 

Here the examples using 256 shapes from pypsa-eur clustering

With `if` 
```python 
%time indicator_matrix = cutout.indicatormatrix(shapes.geometry)
CPU times: user 3.53 s, sys: 3.84 ms, total: 3.54 s
Wall time: 3.53 s
```

Without `if`
```python
%time indicator_matrix = cutout.indicatormatrix(shapes.geometry)
CPU times: user 4.81 s, sys: 4.7 ms, total: 4.82 s
Wall time: 4.81 s
```

FabianHofmann:
> > For chaching properties, [this](https://docs.python.org/3/library/functools.html#functools.cached_property) decorator would be perfect
> 
> Or `CachedAttribute` which has been available in vresutils since 2015. https://github.com/FRESNA/vresutils/blob/5d5fbb5d19454042c5bc1c379b4634d1182cfd6a/vresutils/decorators.py#L187-L208

This is nice! Works as intended. I would copy it to atlite.utils for not having an extra dependency on vresutils. If you oppose please shout out.

PR #72, opened 2020-05-12
sarah: fix file parsing with pandas Period

FabianHofmann:
When creating a cutout for a whole year, with `module='sarah'`, the sarah data is only included for the last month. The bug is caused by a wrong assignment when parsing the raw sarah files. This PR fixes this bug, by searching for all files which begin with 
```python
str(period)
```
in case `period` is a `pandas.Period`


FabianHofmann:
close to due to d2d78074f532dab8b44d46512fbeac702ea76f70

PR #74, opened 2020-05-13
Expand test scripts

FabianHofmann:
(Unfortunately I don't yet see the template yet)

This PR expands the test scripts, by double checking the conversion functions (pv, wind, runoff) against some obvious heuristics. 

Also, I cleaned up some scripts with very long lines (only style correction, no code functionality change).


fneum:
Template will only show if in default branch (i.e. `master`)

euronion:
Running `pytest` results in the following Failure:

```
=========================================================================================================================================================================================================== FAILURES ===========================================================================================================================================================================================================
__________________________________________________________________________________________________________________________________________________________________________________________________ test_dummy_delete_tmp_dir ___________________________________________________________________________________________________________________________________________________________________________________________________

    def test_dummy_delete_tmp_dir():
        '''
        Ignore this test its only purpose is to delete temporary cutout files
        created for the test run
        '''
        for nc in tmp_dir.iterdir():
>           os.remove(nc)
E           OSError: [Errno 26] Text file busy: 'tmp_files_test/era5_test.nc'

test/test_cutout_preparation.py:170: OSError

```

I assume by "Ignore this test" you do not mean that the failure is to be ignored?

Else everything is passing with a few (unwanted) deprecation warnings. I'll handle them in the `v0.2` PR directly.

euronion:
If I understand [this](https://docs.pytest.org/en/latest/tmpdir.html) correctly, `pytest` provides a temporary directory out-of-the-box which then gets cleaned up after the run.

FabianHofmann:
> I assume by "Ignore this test" you do not mean that the failure is to be ignored?

Exactly, I'm not sure why the file is still  busy. It should be closed after the cutout creation right?

> Else everything is passing with a few (unwanted) deprecation warnings. I'll handle them in the `v0.2` PR directly.

Great!

FabianHofmann:
> If I understand [this](https://docs.pytest.org/en/latest/tmpdir.html) correctly, `pytest` provides a temporary directory out-of-the-box which then gets cleaned up after the run.

Yes, but the longer I'm thinking of it, the better it seems to make a proper tmpdir

FabianHofmann:
> If I understand [this](https://docs.pytest.org/en/latest/tmpdir.html) correctly, `pytest` provides a temporary directory out-of-the-box which then gets cleaned up after the run.

This is nice (didn't see that comment before)

FabianHofmann:
@euronion the pytest.fixture is only working for tempfiles within the test_ functions. So I stick to the original structure of having a dummy test which deletes the tmpdir. But now it should now also work for windows machines. Can you have a test on the pytest run?

euronion:
I didn't encounter any problems running it now. Only the **expected** deprecation warnings from `atlite` and `xarray` (which were fixed in the `v0.2` branch).
No errors.

PR #82, opened 2020-06-05
Add travis-ci configuration

coroa:
- Runs tests
- Deploys new tags on master to PyPI

@fabianhofmann:
I activated travis on atlite. So i'd suggest, merging this into master, then rebasing the `v0.0.3` branch ontop of it. Merging the release branch when it is ready to master.

And then going to https://github.com/PyPSA/atlite/releases and "Draft a new release" to generate the new tag and thus push the release to pypi. If that works I'd add the same to vresutils so that all of you should be able to create new releases.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,46 @@
+language: bash
+sudo: false # Use container-based infrastructure
+
+os:
+  - windows
+  - linux
+  - osx
+
+env:
+  - PYTHON_VERSION="3.7"
+  - PYTHON_VERSION="3.8"
+
+before_install:
+  - wget https://raw.githubusercontent.com/trichter/conda4travis/latest/conda4travis.sh -O conda4travis.sh
+  - source conda4travis.sh
+
+install:
+  - conda config --add pinned_packages python=$PYTHON_VERSION
+  - conda create -n atlite python pip wheel pytest
+  - conda env update -n atlite --file=environment.yaml
+  - conda activate atlite
+  # - conda install -q -c conda-forge python-coveralls # don't install on appveyor
+  - pip install --no-cache-dir .
+
+# before_script:  # configure a headless display to test plot generation
+#   - "export DISPLAY=:99.0"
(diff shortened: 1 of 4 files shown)


FabianHofmann:
Sweet!

FabianHofmann:
seems that pytest is missing in the environment

FabianHofmann:
It crosses my mind that one need an cdsapi key for retrieving the era5 data right? This might as well be a problem

fneum:
> It crosses my mind that one need an cdsapi key for retrieving the era5 data right? This might as well be a problem

We can try encrypted variables, but it's differs from the way the CDS API keys are stored in `.cdsapirc`: https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+  - PYTHON_VERSION="3.7"
+  - PYTHON_VERSION="3.8"
+
+before_install:
+  - wget https://raw.githubusercontent.com/trichter/conda4travis/latest/conda4travis.sh -O conda4travis.sh
+  - source conda4travis.sh
+
+install:
+  - conda config --add pinned_packages python=$PYTHON_VERSION
+  - conda create -n atlite python pip wheel
```
fneum, line 18:
```suggestion
  - conda create -n atlite python pip wheel pytest
```
We can try this. I think I'd prefer this over an extra `environment.dev.yaml`.

fneum: Ah, or just add it to the `environment.yaml`.

fneum:
Here's some details on using `cdsapi` with travis: https://github.com/bluegreen-labs/ecmwfr/issues/3

FabianHofmann:
great thanks!

.travis.yml
coroa: > Ah, or just add it to the environment.yaml.

Yes, that was the essence of the third commit.

PR #83, opened 2020-06-06
tests: Use pytest fixtures and marks

coroa:
Has been tested locally.


FabianHofmann:
This is great and works like charm

PR #84, opened 2020-06-07
setup: Use git tag versioning

coroa:
Uses [setuptools_scm](https://github.com/pypa/setuptools_scm) to:
- compute a version number based on the latest git tag
- write it to `atlite/version.py`
- superseed MANIFEST.in (auto-generated by listing git repo)

No need to adapt `_version.py` anymore.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,0 @@
-include atlite/resources/*/*.yaml
-include README.rst LICENSE.txt
-#include requirements.yml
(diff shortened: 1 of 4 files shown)


coroa:
Yes, this is how it is supposed to be. At least more or less.

It knows you are somewhere after v0.0.2, since that's the previous tag. Thus you are now into some development pre-release to v0.0.3, based on the git commit b9a7edd. I don't know, how it determines the 10.

coroa:
Ok, here's the explanation: https://github.com/pypa/setuptools_scm#default-versioning-scheme

FabianHofmann:
that's nice!

Issue #47, opened 2019-09-17
PermissionError during temp file cleanup after cutout download (Windows)


euronion:
Encountering the following error when creating a cutout (ERA5):

```
INFO:cdsapi:Download rate 3.8M/s

[                                        ] | 0% Completed |  0.0s
[########################################] | 100% Completed |  0.3s

---------------------------------------------------------------------------
PermissionError                           Traceback (most recent call last)
<ipython-input-5-3416e809141b> in <module>
      8 
      9 # This is where all the work happens (this can take some time, for us it took ~15 minutes).
---> 10 cutout.prepare()

c:\users\j. hampp\documents\github\atlite\atlite\data.py in cutout_prepare(cutout, features, freq, tmpdir, overwrite)
    226     finally:
    227         if not keep_tmpdir:
--> 228             rmtree(tmpdir)
    229 
    230     # Re-open

C:\anaconda\envs\atlite\lib\shutil.py in rmtree(path, ignore_errors, onerror)
    511             # can't continue even if onerror hook returns
    512             return
--> 513         return _rmtree_unsafe(path, onerror)
    514 
    515 # Allow introspection of whether or not the hardening against symlink

C:\anaconda\envs\atlite\lib\shutil.py in _rmtree_unsafe(path, onerror)
    395                 os.unlink(fullname)
    396             except OSError:
--> 397                 onerror(os.unlink, fullname, sys.exc_info())
    398     try:
    399         os.rmdir(path)

C:\anaconda\envs\atlite\lib\shutil.py in _rmtree_unsafe(path, onerror)
    393         else:
    394             try:
--> 395                 os.unlink(fullname)
    396             except OSError:
    397                 onerror(os.unlink, fullname, sys.exc_info())

PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\J705E~1.HAM\\AppData\\Local\\Temp\\tmplaut1buy\\tmp6jjdud34.nc'
```

* Error seems new, haven't encountered it before.
* The download finishes, but no `cutout.data` data variables are available.
* Probably Windows related (Filehandlers ...)
* Possibly related to the changes made in https://github.com/PyPSA/atlite/commit/9273a15169448e71f9af3a23136f322798a6c982

coroa:
[...]

PR #86, opened 2020-06-07
era5: request only one dataset per feature

FabianHofmann:
As detected by @euronion, the era5 download leads to weird (for me explainable) errors in some cases. This has to do with the delayed concatenation of datasets which are stored in temporary .nc files. This PR solves this issue by downloading only one temporary netcdf file per feature. 

Further advantage: 
* Use cdsapi progressbar to track the download, looks nice and gives a good estimation of remaining downloading time
* the code becomes clearer as we don't have to loop over several request for one variable

### DIFF
--- a/atlite/data.py
+++ b/atlite/data.py
@@ -95,12 +95,6 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
         Cutout with prepared data. The variables are stored in `cutout.data`.
 
     """
-    if tmpdir is None:
-        tmpdir = mkdtemp()
-        keep_tmpdir = False
-    else:
-        keep_tmpdir = True
-
     modules = atleast_1d(cutout.module)
     features = atleast_1d(features)
     prepared = set(cutout.data.attrs['prepared_features'])
@@ -134,9 +128,6 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
                 mode = 'w'
 
             ds.to_netcdf(cutout.path, mode=mode)
-
-    if not keep_tmpdir:
-        rmtree(tmpdir)
 
     cutout.data = xr.open_dataset(cutout.path, chunks=cutout.chunks)
 
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -16,6 +16,7 @@ import numpy as np
 import xarray as xr
 from tempfile import mkstemp
 import cdsapi
+import weakref
 
 from ..gis import maybe_swap_spatial_dims
 
@@ -203,6 +204,17 @@ def retrieval_times(coords, static=False):
                 'time': list(time.strftime("%H:00").unique())}
 
 
+
+def noisy_unlink(path):
+    """Delete file at given path."""
+    logger.info(f"Deleting file {path}")
+    try:
+        os.unlink(path)
+    except PermissionError:
+        logger.error(f"Unable to delete file {path}, as it is still in use.")
+
+
+
 def retrieve_data(product, chunks=None, tmpdir=None, **updates):
     """Download data like ERA5 from the Climate Data Store (CDS)."""
     # Default request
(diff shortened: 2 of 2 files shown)


coroa:
Can we run into a too large file to request error, if you want, say, a 5 year-long european wind cutout?

atlite/datasets/era5.py
```diff
@@ -192,86 +188,39 @@ def _area(coords):
...
 
     fd, target = mkstemp(suffix='.nc', dir=tmpdir)
     os.close(fd)
 
     result.download(target)
 
-    ds = xr.open_dataset(target, chunks=chunks or {})
-    if tmpdir is None:
-        logger.debug(f"Adding finalizer for {target}")
-        weakref.finalize(ds._file_obj._manager, noisy_unlink, target)
```
coroa, line 273: Why do you remove the automatic deletion?

atlite/datasets/era5.py
```diff
@@ -192,86 +188,39 @@ def _area(coords):
...
 
-    result = cdsapi.Client(progress=False).retrieve(product, request)
+    result = cdsapi.Client().retrieve(product, request)
 
     fd, target = mkstemp(suffix='.nc', dir=tmpdir)
     os.close(fd)
 
     result.download(target)
 
-    ds = xr.open_dataset(target, chunks=chunks or {})
```
coroa, line 270: You are removing the `or {}` part from this line. It's functionality was to always load the dataset with dask even if one did not provide any chunks. Now it will load the dataset directly with NetCDF5. Is this intentional?

atlite/datasets/era5.py
FabianHofmann: You're right. I was expecting that tmpdir is always not `None`. In previous versions of `v0.2` there was an `assert tmpdir not None`. It's not there anymore so the deletion makes sense again. I'll revert it.

atlite/datasets/era5.py
FabianHofmann: Yes it is. When setting the chunks to `None` in the cutout creation, the process should work without `dask` at every point in the process, the `or {}` overwrites that. We can set the default of this argument to `{}`, but we should leave the possibility open to deactivate `dask`.

coroa: Just checking, sounds sensible

FabianHofmann:
> Can we run into a too large file to request sort of error, if you want, say, a 5 year-long european wind cutout?

Good point, cdsapi allows 120000 items per request. That's roughly a cutout for European bounds and 4 years. Should we cover this case, or leave it to the user to build multiple cutouts then?

coroa:
> > Can we run into a too large file to request sort of error, if you want, say, a 5 year-long european wind cutout?
> 
> Good point, cdsapi allows 120000 items per request. That's roughly a cutout for European bounds and 4 years. Should we cover this case, or leave it to the user to build multiple cutouts then?

I don't know. I'd like to not have to tell anyone he should split cutouts him/herself if it's easy for us to do. What is so bad about the delayed merge? We could for a simplification just not use the dask delayed feature and download one cutout after the other. Would that help?

atlite/data.py
```diff
@@ -135,9 +129,6 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
 
             ds.to_netcdf(cutout.path, mode=mode)
 
-    if not keep_tmpdir:
-        rmtree(tmpdir)
-
```
FabianHofmann, line 140: As we have an automated deletion of the tmp files in `era5.py` and all other modules (`sarah`, `gebco`) don't produce any tempfiles, I removed the tmpdir creation and deletion. If tmpdir is None, the datasets modules should delete any produced tmp files automatically.  If tmpdir is specified, the files will be kept.

coroa: Not a good idea! Have been down that road before.

> A direct consequence for making the cutout preparation work for sarah was to let go of the cleanup tmp files automatically once the file handle is released (the weakref.finalizer(noisy_unlink, ...) bit), since dask is often configured to use multiple processes and then whether a file handle is released in one process is not a good indicator for whether we still need the tmp file. Thus, I was forced to switch to a create all files in a temporary directory and clean up the directory when the cutout is ready scheme.
from https://github.com/PyPSA/atlite/issues/47#issuecomment-532579112

PR #87, opened 2020-06-07
data: Re-enable parallelized queueing

coroa:
Since queueing time is dominant for downloading from ERA-5, I'd propose
to at least have some parallelized support for queueing.

This PR ...
- fixes two bugs with `era5.retrieval_times`
- re-enables warnings
- adds per-dataset dask.delayed concurrency
- introduces a lock to keep progress bars from breaking one another

Would be an alternative for #86 . But might not fix the original reason for it.

### DIFF
--- a/atlite/data.py
+++ b/atlite/data.py
@@ -9,6 +9,9 @@ import xarray as xr
 from numpy import atleast_1d
 from tempfile import mkdtemp
 from shutil import rmtree
+import dask
+from dask.delayed import Delayed
+from dask.utils import SerializableLock
 from dask.diagnostics import ProgressBar
 import logging
 logger = logging.getLogger(__name__)
@@ -21,15 +24,19 @@ def get_features(cutout, module, features, tmpdir=None):
     Load the feature data for a given module.
 
     This get the data for a set of features from a module. All modules in
-    `atlite.datasets` all allowed.
+    `atlite.datasets` are allowed.
     """
     parameters = cutout.data.attrs
+    lock = SerializableLock()
     datasets = []
     get_data = datamodules[module].get_data
 
     for feature in features:
-        feature_data = get_data(cutout, feature, tmpdir=tmpdir, **parameters)
+        feature_data = get_data(cutout, feature, tmpdir=tmpdir, lock=lock, **parameters)
         datasets.append(feature_data)
+
+    if len(datasets) >= 1 and isinstance(datasets[0], Delayed):
+        datasets = dask.compute(*datasets)
 
     ds = xr.merge(datasets, compat='equals')
     for v in ds:
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -210,17 +210,17 @@ def retrieval_times(coords):
     time = pd.Series(coords['time'])
     time_span = time.iloc[-1] - time.iloc[0]
     if len(time) == 1:
-        return [{'year': d.year, 'month': d.month, 'day': d.day,
+        return [{'year': str(d.year), 'month': str(d.month), 'day': str(d.day),
                  'time': d.strftime("%H:00")} for d in time]
     if time_span.days <= 10:
-        return [{'year': d.year, 'month': d.month, 'day': d.day}
+        return [{'year': str(d.year), 'month': str(d.month), 'day': str(d.day)}
                 for d in time.dt.date.unique()]
     elif time_span.days < 90:
-        return [{'year': year, 'month': month}
+        return [{'year': str(year), 'month': str(month)}
                 for month in time.dt.month.unique()
                 for year in time.dt.year.unique()]
     else:
-        return [{'year': year} for year in time.dt.year.unique()]
+        return [{'year': str(year)} for year in time.dt.year.unique()]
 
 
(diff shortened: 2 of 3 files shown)


atlite/datasets/era5.py
```diff
@@ -27,12 +28,6 @@
 import logging
 logger = logging.getLogger(__name__)
 
-import urllib3
-logger.warning('Disable urllib3 and cdsapi warnings.')
-urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
-logging.getLogger("cdsapi").setLevel(logging.ERROR)
```
coroa, line 33: What was the reason for disabling those?

atlite/datasets/era5.py
```diff
@@ -213,19 +208,19 @@ def retrieval_times(coords):
...
+        return [{'year': str(d.year), 'month': str(d.month), 'day': str(d.day)}
                 for d in time.dt.date.unique()]
     elif time_span.days < 90:
-        return [{'year': year, 'month': month}
+        return [{'year': str(year), 'month': str(month)}
                 for month in time.dt.month.unique()
                 for year in time.dt.year.unique()]
     else:
-        return [{'year': year} for year in time.dt.year.unique()]
+        return [{'year': str(year)} for year in time.dt.year.unique()]
```
coroa, line 223: if i don't serialize the integers to strings here, I get a JSON serialization error: Can't serialize int64.

atlite/datasets/era5.py
FabianHofmann: I was annoyed by the continuous request info and `urllib3` spit out a lot of warning, but apparently the latter was caused by a wrong verification setup in my personal properties

atlite/datasets/era5.py
FabianHofmann: right

atlite/datasets/era5.py
```diff
@@ -213,19 +208,19 @@ def retrieval_times(coords):
 
     """
     time = pd.Series(coords['time'])
-    time_span = time[0] - time[len(time)-1]
+    time_span = time.iloc[-1] - time.iloc[0]
     if len(time) == 1:
```
FabianHofmann, line 212: That was an unlucky bit, just recognized that part yesterday.

atlite/datasets/era5.py
```diff
@@ -307,24 +308,25 @@ def get_data(cutout, feature, tmpdir, **creation_parameters):
...
-        res = dask.compute(*datasets)
-    return xr.concat(res, dim='time')
+        return ds
+
+    if feature in static_features:
+        return retrieve_once(retrieval_times(coords)[0])
+
+    datasets = map(retrieve_once, retrieval_times(coords))
+
+    return delayed(xr.concat)(datasets, dim='time')
```
FabianHofmann, line 332: See above, now the `get_data` functions should have a standard output format.

atlite/data.py
```diff
@@ -21,16 +24,20 @@ def get_features(cutout, module, features, tmpdir=None):
...
     """
     parameters = cutout.data.attrs
+    lock = SerializableLock()
     datasets = []
     get_data = datamodules[module].get_data
 
     for feature in features:
-        feature_data = get_data(cutout, feature, tmpdir=tmpdir, **parameters)
+        feature_data = get_data(cutout, feature, tmpdir=tmpdir, lock=lock, **parameters)
         datasets.append(feature_data)
```
FabianHofmann, line 36: The expected output of all `get_data` functions should be the same, always a `xr.Dataset`, not delayed objects. What speaks against adding the delay here? (instead of within `era5.py`)
`delayed(get_data)(cutout, feature, tmpdir=tmpdir, lock=lock, **parameters)`
Then we'd also have a parallel queuing

atlite/data.py
```diff
@@ -21,16 +24,20 @@ def get_features(cutout, module, features, tmpdir=None):
...
 
+    if len(datasets) >= 1 and isinstance(datasets[0], Delayed):
+        datasets = dask.compute(*datasets)
```
FabianHofmann, line 39:
```suggestion
        datasets, = dask.compute(*datasets)
```
This lead to an error as `dask.compute` returns a tuple

coroa: Doesn't lead to an error for me. Datasets are multiple datasets before and after.

atlite/data.py
coroa: Not for multiple years in the same dataset. Maybe that's not soo much of a concern, since the standard is a single download per feature, i guess.

atlite/datasets/era5.py
coroa: I think it would be nice to add two more logging statements of what is requested and what is downloaded to the `retrieve` function and then set the "cdsapi" log level to `logging.WARNING`.

FabianHofmann:
The code works fine and gives the correct results. However when setting two years my waiter does not respond. Working example

```python 
cutout = atlite.Cutout(path, chunks={'time':100},
                        x=slice(-0.,3.), y=slice(50.,52.),
                        time=slice('2013-01', '2015'), module='era5')
cutout.prepare(features = ['temperature'])
```

PR #88, opened 2020-06-08
gis: Fix rasterio crs specification

coroa:


euronion:
How did this change become necessary?
Does this change introduce a minimum version dependency on `rasterio`?

euronion:
The docstring also needs an update:
```
      - src_crs, dst_crs define the different crs (default: latlong)
```

coroa:
> How did this change become necessary?

Yes, an update to rasterio, i guess.

> Does this change introduce a minimum version dependency on `rasterio`?

Probably.

coroa:
And probably this change to rasterio is more than 10months back, because i made a similar change in the v0.2 branch then at https://github.com/PyPSA/atlite/commit/507b59e6cd50299b2b5d5aa9efb775a694134ab0 .

Can't research the exact version constraints now.

euronion:
I guess it's fine then. No need to waste time on pinpointing the breaking release of `rasterio`  then.

coroa:
Seems to work with 1.0. That's old enough.

PR #91, opened 2020-06-08
sanitize static variables in era5

### LINKED ISSUE #89: era5 time coordinates for static variables
There is a small bug when loading the static variable only, i.e. 'height' with era5. This will have the time dimension of the first time parameters of the first request. I'll suggest a solution

FabianHofmann:
closes #89

### DIFF
--- a/atlite/data.py
+++ b/atlite/data.py
@@ -9,8 +9,7 @@ import xarray as xr
 from numpy import atleast_1d
 from tempfile import mkdtemp
 from shutil import rmtree
-import dask
-from dask.delayed import Delayed
+from dask import delayed, compute
 from dask.utils import SerializableLock
 from dask.diagnostics import ProgressBar
 import logging
@@ -32,11 +31,11 @@ def get_features(cutout, module, features, tmpdir=None):
     get_data = datamodules[module].get_data
 
     for feature in features:
-        feature_data = get_data(cutout, feature, tmpdir=tmpdir, lock=lock, **parameters)
+        feature_data = delayed(get_data)(cutout, feature, tmpdir=tmpdir,
+                                         lock=lock, **parameters)
         datasets.append(feature_data)
 
-    if len(datasets) >= 1 and isinstance(datasets[0], Delayed):
-        datasets = dask.compute(*datasets)
+    datasets = compute(*datasets)
 
     ds = xr.merge(datasets, compat='equals')
     for v in ds:
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -15,18 +15,15 @@ import os
 import pandas as pd
 import numpy as np
 import xarray as xr
-import dask
-from dask import delayed
-from dask.utils import SerializableLock
 from tempfile import mkstemp
 import weakref
 import cdsapi
 
 from ..gis import maybe_swap_spatial_dims
 
-
 import logging
 logger = logging.getLogger(__name__)
+logging.getLogger("cdsapi").setLevel(logging.ERROR)
 
 # Model and Projection Settings
 projection = 'latlong'
@@ -187,7 +184,7 @@ def _area(coords):
     return [y1, x0, y0, x1]
 
 
-def retrieval_times(coords):
+def retrieval_times(coords, static=False):
     """
     Get list of retrieval cdsapi arguments for time dimension in coordinates.
 
@@ -208,7 +205,12 @@ def retrieval_times(coords):
 
     """
     time = pd.Series(coords['time'])
+    if static:
+        return {'year': str(time[0].year), 'month': str(time[0].month),
+                'day': str(time[0].day), 'time': time[0].strftime("%H:00")}
+
     time_span = time.iloc[-1] - time.iloc[0]
+
     if len(time) == 1:
         return [{'year': str(d.year), 'month': str(d.month), 'day': str(d.day),
                  'time': d.strftime("%H:00")} for d in time]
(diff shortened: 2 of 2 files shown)


atlite/data.py
```diff
@@ -124,28 +128,29 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
...
 
+        ds = cutout.data.merge(ds[missing_vars.values])\
+                            .assign_attrs(**cutout.data.attrs, **ds.attrs)
+        ds.attrs = non_bool_dict(cutout.data.attrs)
+        ds.attrs.update({'prepared_features': list(prepared)})
+
+        # write data to tmp file, copy it to original data, this is much saver
+        # than appending variables
+        tstr = next(tmpname())
+        tmp = cutout.path.parent / (f'tmp_{tstr}' + cutout.path.stem + '.nc')
```
coroa: It would be better to use
```suggestion
        tmp = mktemp(prefix="tmp_", suffix=f"_{cutout.path.stem}.nc", dir=cutout.path.parent)
```
since it accepts only candidates which do not exist yet.

If you are worried by the note in the doc-string:

> THIS FUNCTION IS UNSAFE AND SHOULD NOT BE USED.  The file name may
refer to a file that did not exist at some point, but by the time
you get around to creating it, someone else may have beaten you to
the punch.

Then your best option is to take what I used here before:
https://github.com/PyPSA/atlite/blob/09714d595fc09f2aa1ba79bdd9e60391c582e880/atlite/data.py#L211-L215

euronion, line 141:
```suggestion
        tmp = cutout.path.parent / f'tmp_{tstr}_{cutout.path.stem}.nc'
```

euronion:
I don't understand much of it, from the first glance I couldn't find anything disturbing.

Better have @coroa look at it with his eagle eyes!

atlite/data.py
```diff
@@ -124,28 +128,29 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
...
         with ProgressBar():
-            if cutout.path.exists():
-                mode = 'a'
-            else:
-                mode = 'w'
+            ds.to_netcdf(tmp)
+            ds.close()
+        cutout.data.close()
```
coroa, line 145: If this is a user provided dataset, i don't think we should close it. That's what the `is_view` attribute was for.

FabianHofmann:
That's fine, but thanks for having a look!

test/test_preparation_and_conversion.py
```diff
@@ -37,6 +38,15 @@ def prepared_features_test(cutout):
     assert set(cutout.prepared_features) == set(cutout.data)
 
 
+def update_feature_test(cutout):
+    """atlite should be able to overwrite a feature."""
+    tmp = Path(mkstemp(suffix='.nc')[1])
+    tmp.unlink()
```
coroa, line 44: This probably breaks on a windows installation. You are creating a new temp file for writing, dropping its file handle (effectively leaking it, since it is a lowlevel file handle the file will not be closed until shutting down the python interpreter), then unlinking it (re-creating the race-condition, which the doc-string of `mktemp` warns about), ie deleting an opened file, what windows screams about (maybe it does not since it's within the same process).

The only safe way i found is the one from
https://github.com/PyPSA/atlite/blob/09714d595fc09f2aa1ba79bdd9e60391c582e880/atlite/data.py#L211-L215

Here i would just use the `tmp_path` fixture: https://docs.pytest.org/en/latest/tmpdir.html#the-tmp-path-fixture

atlite/data.py
```diff
@@ -32,11 +31,11 @@ def get_features(cutout, module, features, tmpdir=None):
     get_data = datamodules[module].get_data
 
     for feature in features:
-        feature_data = get_data(cutout, feature, tmpdir=tmpdir, lock=lock, **parameters)
+        feature_data = delayed(get_data)(cutout, feature, tmpdir=tmpdir,
+                                         lock=lock, **parameters)
```
coroa, line 35: Did you test, whether this works to queue the requests in parallel?

atlite/data.py
```diff
@@ -124,28 +128,28 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
...
+        cutout.data.close()
+        if cutout.path.exists():
+            cutout.path.unlink()
+        tmp.rename(cutout.path)
 
-            ds.to_netcdf(cutout.path, mode=mode)
+        cutout.data = xr.open_dataset(cutout.path, chunks=cutout.chunks)
 
     if not keep_tmpdir:
         rmtree(tmpdir)
```
coroa, line 153: `tmpdir` cleanup should be in a `finally`-close or an exception will leak the tmpdir.

atlite/data.py
```diff
@@ -124,28 +128,28 @@ def cutout_prepare(cutout, features=slice(None), tmpdir=None, overwrite=False):
...
 
+        ds = cutout.data.merge(ds[missing_vars.values])\
+                            .assign_attrs(**cutout.data.attrs, **ds.attrs)
+        ds.attrs = non_bool_dict(cutout.data.attrs)
+        ds.attrs.update({'prepared_features': list(prepared)})
+
+        # write data to tmp file, copy it to original data, this is much saver
+        # than appending variables
+        tstr = next(tmpname())
+        tmp = cutout.path.parent / f'tmp_{tstr}_{cutout.path.stem}.nc'
```
coroa, line 141: Please use `mktemp` instead of the private `_get_candidate_names`, since it also makes sure the candidate does not exist yet:
```suggestion
        # write data to tmp file, copy it to original data, this is much safer
        # than appending variables
        tmp = Path(mktemp(suffix=f"_{cutout.path.stem}.nc, dir=cutout.path.parent))
```

As the next improvement step you should worry about the race condition described in the doc-string of `mktemp`:
> THIS FUNCTION IS UNSAFE AND SHOULD NOT BE USED.  The file name may refer to a file that did not exist at some point, but by the time you get around to creating it, someone else may have beaten you to the punch.

And decide whether you want to use instead the pattern introduced in
https://github.com/PyPSA/atlite/blob/09714d595fc09f2aa1ba79bdd9e60391c582e880/atlite/data.py#L211-L215

atlite/datasets/era5.py
```diff
@@ -15,18 +15,15 @@
...
 from tempfile import mkstemp
 import weakref
 import cdsapi
 
 from ..gis import maybe_swap_spatial_dims
 
-
 import logging
 logger = logging.getLogger(__name__)
+logging.getLogger("cdsapi").setLevel(logging.ERROR)
```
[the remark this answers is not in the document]
coroa: See #94

PR #92, opened 2020-06-09
Update to pyproj 2

coroa:

- Following the lead of geopandas, pyproj, cartopy and rasterio, we use `crs` to indicate what we meant by a projection up to now.
~- New grid_cell_areas CachedAttribute to simplify calculation of areas~

## How Has This Been Tested?

Hasn't been tested yet

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist


FabianHofmann:
this is great! 
A question about the area. Why not using a geopandas built-in? Somthing like 
```python 
areakm = gpd.GeoSeries(cutout.grid_cells, crs=cutout.crs)\
              .to_crs({'proj': 'cea'}).area
```
(derived together from https://gis.stackexchange.com/questions/218450/getting-polygon-areas-using-geopandas)

atlite/gis.py
```diff
@@ -100,6 +101,23 @@ def reproject(shapes, p1, p2):
...
 
+def grid_cell_areas(cutout):
+    """
+    Compute the area of each grid cell in km2
+    """
+    if cutout.crs == CRS.from_epsg(4326):
+        # use equation derived in https://www.pmel.noaa.gov/maillists/tmap/ferret_users/fu_2004/msg00023.html
+        y = cutout.coords['y']
+        dy = cutout.dy
+        area_km2 = ((2 * np.pi * R**2) * (dx / 360) *
```
coroa, line 112: ups, `R` is the earth radius in km. Should add that constant somewhere.

coroa:
> this is great!
> A question about the area. Why not using a geopandas built-in? Somthing like
> 
> ```python
> areakm = gpd.GeoSeries(cutout.grid_cells, crs=cutout.crs)\
>               .to_crs({'proj': 'cea'}).area
> ```
> 
> (derived together from https://gis.stackexchange.com/questions/218450/getting-polygon-areas-using-geopandas)

Yes, would also work. I was afraid of using yet another concept for `crs` definitions. I am not sure they are compatible with the `pyproj.CRS` ones. Also i decided for [EPSG:3035](https://epsg.io/3035) as an equal area projection since it is well tailored to europe, but then i don't know `'cea'`.

euronion:
> Also i decided for [EPSG:3035](https://epsg.io/3035) as an equal area projection since it is well tailored to europe, but then i don't know `'cea'`.

What happens if the region is outside of Europe / the WGS84 bounds mentioned in the link?
Do we then loose on accuracy (significant or acceptable)? Or do we get an error?

coroa:
> > Also i decided for [EPSG:3035](https://epsg.io/3035) as an equal area projection since it is well tailored to europe, but then i don't know `'cea'`.
> 
> What happens if the region is outside of Europe / the WGS84 bounds mentioned in the link?
> Do we then loose on accuracy (significant or acceptable)? Or do we get an error?

We loose on accuracy, don't know by how much.

euronion:
> > > Also i decided for [EPSG:3035](https://epsg.io/3035) as an equal area projection since it is well tailored to europe, but then i don't know `'cea'`.
> > 
> > 
> > What happens if the region is outside of Europe / the WGS84 bounds mentioned in the link?
> > Do we then loose on accuracy (significant or acceptable)? Or do we get an error?
> 
> We loose on accuracy, don't know by how much.

I find it odd to use a projection focus on Europe here. `cae` probably refers to [Equal Area Cylindrical](https://proj.org/operations/projections/cea.html) w/ a world-wide scope. If it also works with `cae` or any other ww-defined equal area projection I would favour any of those over EPSG:3035.

PR #94, opened 2020-06-09
era5: Logging differentiates between request and download

coroa:


FabianHofmann:
I'm not sure if this is a good idea, it really blows up when you prepare a cutout for 5 days. 

Why not being happy with the option to enable the full log with logging.basicConfig(level=logging.DEBUG)? This what I made the `info_callback` thing for. And with the progressbar enable (in #91 it's currently not working, but there is a small fix for it) you have a good insight where the download is.

coroa:
If you prepare a cutout for a year on the other hand, the preparation on the CDS site takes sometimes up to 15min, during which you don't have any indication of progress. I actually killed it a couple of times, because i was unsure, whether it froze. And then during the downloads you didn't know what it was downloading.

PR #95, opened 2020-06-10
[v0.2] review suggestions

coroa:


FabianHofmann:
Please always run pytest (as long as travis is not activated). Prevents a lot of 'unneeded' review

PR #104, opened 2020-11-09
add support for ERA5 back extension to 1950

fneum:
The recent back extension of the ERA5 from 1950-1978 has a different product name. This is accounted for in this PR.

https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels-preliminary-back-extension?tab=overview

### DIFF
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -124,6 +124,9 @@ def _rename_and_clean_coords(ds, add_lon_lat=True):
     return ds
 
 def prepare_meta_era5(xs, ys, year, month, module):
+    product = 'reanalysis-era5-single-levels'
+    if int(year) < 1979: product += '-preliminary-back-extension'
+    
     # Reference of the quantities
     # https://confluence.ecmwf.int/display/CKB/ERA5+data+documentation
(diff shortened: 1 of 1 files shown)


euronion:
Do you know whether the function for multi-year cutouts is called once or multiple times per year or only once for the full cutout period?
In the first case the PR seems fine, in the second case the would fail for cutouts including years <1979 and >= 1979.

We'll have to check `v0.2` at some point whether it works there as well. (at some time in the future).

euronion:
(+ According to schedule: around the end of 2021 the preliminary version will be integrated into the regular dataset and the preliminary version become deprecated)

fneum:
> Do you know whether the function for multi-year cutouts is called once or multiple times per year or only once for the full cutout period?

There is an API request for each (year, month) tuple. So no conflict AFAICS.

> We'll have to check v0.2 at some point whether it works there as well.

Yes, for another time.

>  around the end of 2021 the preliminary version will be integrated into the regular dataset

Yes, long enough to include the preliminary version now IMO.

euronion:
Ok / ACK / Agreed. PR is G2G then.

PR #106, opened 2020-11-12
Support negative latitudes to PV panel orientation

### LINKED ISSUE #103: Fail to convert weather data to solar power for sites with latitude < 0
An error, saying 'Not implemented for negative latitudes', was raised when I converted weather data in regions with negative latitudes. 
Can Atlite still get the right converted results after the 'abs()' function is added to 'orientation.py'? Cheers.

euronion:
Fixes #103 .

Untested yet. Await response by @WeselyDAN .


FabianHofmann:
I tested it and compared it to renewable ninjas. Everything thing works as expected

PR #110, opened 2021-01-12
Ci mamba

FabianHofmann:
Use mamba for CI on travis, reduces time tremendously

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,7 @@ before_install:
   - source conda4travis.sh
 
 install:
-  - conda config --set remote_max_retries 5
-  - conda config --set channel_priority strict
-  - conda install mamba=0.1.2 -c conda-forge -y # fix the version for faster installation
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -11,11 +11,10 @@ before_install:
   - source conda4travis.sh
 
 install:
-  - conda config --set remote_max_retries 5
-  - conda config --set channel_priority strict
-  - conda create -n atlite python pip pytest
-  - conda env update -n atlite --file=environment.yaml
+  - conda install mamba=0.1.2 -c conda-forge # fix the version for faster installation
```
euronion, line 14: I don't like fixing the version number, we would have to update this continously and might install non-recent packages.
Is this a necessary limitation (performance-wise) or would a less strict statement like  `mamba>=0.1.2` yield a similar improvement? @FabianHofmann

FabianHofmann: Actually you're right! The improvement is sufficiently small. I'll remove the fixing

PR #111, opened 2021-01-12
fix numpy <-> daks incompatibility in clip function

FabianHofmann:


euronion:
Can you add a comment to the code referencing the incompatibility (issue or documentation) so that one doesn't change it unintentionally?

FabianHofmann:
refering to https://github.com/pydata/xarray/issues/4796#issuecomment-758563789

PR #153, opened 2021-06-08
Introduce black coding style and pre-commit hooks for black 

FabianHofmann:

As we discussed some times, we should add a pre-commit hook for users to be aligned with the linting. This PR aligns all code to black linting style (brute force) and adds a pre-commit hook for black (and check yaml files)
 
## Note!
No content change, only style. Except for environment.yaml, setup.py and .pre-commit-config.yaml

## Motivation and Context
No need to check code style in PRs anymore. 

## Type of change

## Checklist


euronion:
I find the idea of running `pre-commit` locally attractive. At the same time I am working on something similar but using GitHub Actions such that it is run on every PR rather than on every commit. See https://github.com/PyPSA/pypsa-eur/pull/242 .

I'll see how I can add `black` and other linters/formatters.

Then either we consider our two approaches competing and only choose one. Or could they be complementary (`pre-commit` hooks for changing, GH actions for checking&nagging)? What do you think?

euronion:
Alright, I am liking this idea more and more! :)

This commit will change a lot of things due to `black` magic. So for future reference:

**Running `git blame` will often return this PR. To avoid this behaviour use**
```
git blame --ignore-rev <HASH>
```
**where <HASH> is the hash of this PR's commit. This will ignore this commit with it's `black` changes in the `git blame` command.**

I'll push a few updated/changes to the branch, hope you don't mind.

Furthermore:
I think we should use `environment.yaml` for setting up the `dev` environment / people using GitHub for installing the package. `setup.py` should only be "for users", i.e. remove `black` and `pre-commit` from the dependencies there. What do you think?

setup.py
```diff
@@ -7,51 +7,55 @@
...
+        "pyyaml",
+        "rasterio>=1.0",
+        "shapely",
+        "progressbar2",
+        "tqdm",
+        "pyproj>=2",
+        "geopandas",
+        "cdsapi",
+        "pytest",
+        "black",
```
euronion, line 47: `black` is not a necessary dependence if only used in `pre-commit`.

setup.py
```diff
@@ -7,51 +7,55 @@
...
+        "pre-commit",
```
euronion, line 48: I think we should have `pre-commit` not as a package dependency, but only as a development requirement, e.g. only in `environment.yaml` if people install it from GitHub.

environment.yaml
```diff
@@ -27,6 +27,10 @@ dependencies:
   - progressbar2
   - tqdm
 
+  # dev tools
+  - black
+  - pre-commit
```
euronion, line 32: See notes to `setup.py`:
- `black` not required
- `pre-commit` here (`environment.yaml` = development environment)?

setup.py
FabianHofmann: indeed!

setup.py
FabianHofmann: alright let's kick it out then. I think in the long term we should have a some sort of clear dependencies patterns based on the purpose, like having a setup.cfg with sections for **doc**, **dev** etc. Probably this will work well with the `pyproject`

FabianHofmann: I just saw that we already have the dependency spec **doc** in the `setup.py`, we could add **dev** there and add pre-commit there, what you think?

FabianHofmann: (which would probably also the better spot for pytest)

euronion: Yes, I think that's a good idea. Call it `dev`?. Then it's `pip install atlite[dev]`. Does `conda-forge` replicate the `extra_requires`?

`setup.py` --> normal use/doc/dev
`environment.yaml` -> = dev from setup.py

setup.py
```diff
@@ -43,12 +43,10 @@
...
         "geopandas",
         "cdsapi",
-        "pytest",
-        "black",
-        "pre-commit",
     ],
     extras_require={
-        "docs": ["numpydoc", "sphinx", "sphinx_rtd_theme", "nbsphinx", "nbsphinx-link"]
+        "docs": ["numpydoc", "sphinx", "sphinx_rtd_theme", "nbsphinx", "nbsphinx-link"],
+        "dev": ["black", "pre-commit", "pytest"],
```
euronion, line 49:
```suggestion
        "dev": ["pre-commit", "pytest"],
```

`black` should not be required as dependency (`pre-commit` takes care of black magic on its own).

FabianHofmann: correct!

fneum:
LGTM. One thing we should do after merging (with squash!):

https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame

put full commit hash into a new root level file `.git-blame-ignore-revs`

User has to run one-time locally (don't think you can apply a git config automatically):

```sh
git config blame.ignoreRevsFile .git-blame-ignore-revs
```

OR each time (as @euronion mentioned)

```sh
git blame --ignore-rev <HASH>
```

FabianHofmann:
yes sounds good! @euronion you still want to change something in this PR, or is the current setup fine?

euronion:
Nothing more to add from my side. Green light for the PR.

@fneum I don't think this can be done automatically. Otherwise you could do funny things with other peoples config simply by having them clone a repo.

fneum:
Forgot one thing: [Code style: black](https://github.com/psf/black)

FabianHofmann:
true, can you add that and push it ?

Issue #157, opened 2021-06-13
[JOSS] "community profile" checklist is incomplete


kanderso-nrel:
GitHub has a checklist of repository-level community standards.  Here is the page for atlite: https://github.com/PyPSA/atlite/community

Of the items currently missing, I think the `Contributing` item is most important.  Documenting project conventions like code style rules (pep8, e.g.), testing tools etc would be helpful for new contributors. 

xref https://github.com/openjournals/joss-reviews/issues/3294

FabianHofmann:
@kanderso-nrel we added a CONTRIBUTING.md which directs to the respective part in the documentation. A code of conduct did not seem necessary to us, so far. Please reopen if you still miss something

PR #159, opened 2021-06-14
Create CODE_OF_CONDUCT.md

FabianHofmann:
Addresses #157

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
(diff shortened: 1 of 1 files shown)


euronion:
What about : https://atlite.readthedocs.io/en/latest/contributing.html ?

I don't see much of a content overlap, but we should not have two cages for the one bird.

What I think we could do:
- Link to the "How to contribute" in the `Readme.md` and in the Issue/PR template
- Combine your PR w/ the `Contributing` in RTD
- I'll have a look at the code style issue (related to #153 )

What do you think?

FabianHofmann:
But this is another topic right? The code of conduct is rather a general behavior guideline. In https://github.com/PyPSA/atlite/community it is listed separately from the contributing guideline (this is where I got the idea from). I mean we don't have to include it at all, if there is no need (good-practise packages like [xarray](https://github.com/pydata/xarray) have it though)

euronion:
Short/formal remarks:
- For me "How to contribute" and "how to behave" are closely related in our case.
- IMHO it should in any case be part of the RTD / linked in the contributions page.
- License missing (for REUSE compliance / license by Contributor Convenant Code of Conduct, i.e. CC-BY-4.0 Coraline Ada Ehmke, Copyright 2014)

Longer remarks:
- I never thought of including a Code of Conduct until now. I like the principle idea and wonder whether this COC fits our bill (maybe a shorter version of the same would fit better or a different one?)
- Some parts of the COC do not seem to fit: Like enforcement guidelines, Scope or Enforcement Responsibilities?
- Under "Enforcement" the original has a `[INSERT CONTACT METHOD]` which is missing in your PR and would need to be filled.

FabianHofmann:
I definitely agree with your points. I close this PR and as soon as we see real need for it, we reopen modify...

Issue #175, opened 2021-06-23
Cutout.prepare() never completes


Ovewh:
Cutout preparation ERA5 never completes.

## Description
```
cutout = atlite.Cutout(path="europe_2011_1.nc", module='era5',
                       x=slice(-13,45),
                      y=slice(32,83),
                      time='2011-01-01')    
cutout.prepare() # Never completes
```
## Expected Behavior
Download data and return prepared cutout

## Actual Behavior
The prepare never finishes even though the ERA5 data get downloaded.
## Error Message
I interrupted the program after ~20 minutes
```
2021-06-23 14:00:19,649 INFO Requesting data for feature wind...
2021-06-23 14:00:19,651 INFO Requesting data for feature influx...
2021-06-23 14:00:19,666 INFO NumExpr defaulting to 8 threads.
2021-06-23 14:00:19,913 INFO CDS: Downloading variables
	 * 2m_temperature (2011)
	 * soil_temperature_level_4 (2011)

2021-06-23 14:00:21,616 INFO CDS: Downloading variables
	 * runoff (2011)

2021-06-23 14:00:22,981 INFO CDS: Downloading variables
	 * surface_net_solar_radiation (2011)
	 * surface_solar_radiation_downwards (2011)
	 * toa_incident_solar_radiation (2011)
	 * total_sky_direct_solar_radiation_at_surface (2011)

2021-06-23 14:00:26,740 INFO CDS: Downloading variables
	 * 100m_u_component_of_wind (2011)
	 * 100m_v_component_of_wind (2011)
	 * forecast_surface_roughness (2011)

                                                    

---------------------------------------------------------------------------
KeyboardInterrupt                         Traceback (most recent call last)
<ipython-input-2-eee9441f41e9> in <module>
----> 1 cutout.prepare()

~/miniconda3/envs/atlite_dev/lib/python3.9/site-packages/atlite/data.py in wrapper(*args, **kwargs)
    100             kwargs["tmpdir"] = mkdtemp()
    101             try:
--> 102                 res = func(*args, **kwargs)
    103             finally:
    104                 rmtree(kwargs["tmpdir"])

~/miniconda3/envs/atlite_dev/lib/python3.9/site-packages/atlite/data.py in cutout_prepare(cutout, features, tmpdir, overwrite)
    162         logger.info(f"Calculating and writing with module {module}:")
    163         missing_features = missing_vars.index.unique("feature")
[...]

PR #179, opened 2021-06-28
add notice about processing time CDS

Ovewh:
Added a notice about the  CDS side processing time in the create_cutout example including a link where to 
check the status of CDS request. To inform about why it might take a while before the download starts.

## Motivation and Context
Motivated by #175


euronion:
Thanks, that's a helpful link!

I've just restored the original output from the code cell, such that the output stays displayed in the documentation.

PR #194, opened 2021-11-16
Add CSP modelling capabilities.

euronion:
Closes # (if applicable).

## Type of change

- [n/a] Bug fix (non-breaking change which fixes an issue)
- [n/a] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [n/a] I have added newly introduced dependencies to `environment.yaml` file.


euronion:
@FabianHofmann could you have a look? With a focus on the code.
@pz-max could you have a look? With a focus on the example notebook.

FabianHofmann (commented):
This is really impressive work! The example is excellent and explains well. 

Just a few conceptional questions:
1. I am wondering why the efficiency map is not defined for a azimuth range from 0 to 360. The ranges 0-70 (roughly) and 270 - 360 are missing. Does that mean a solar position in the north is not covered/supported? 
2. I am wondering whether the underestimated efficiencies for low solar altitudes play a role. These have a strong relative deviation from the SAM data. For example: For SAM the combination of altitude=8, azimuth=280 leads to en efficiency of ~10%. The same combination on the derived efficiency map gives  efficiency ~0%. Do you see any harmful potential there? 
3. The underestimated efficiencies for low solar altitudes stand in contrast to the overestimated power outputs in winter. In the example the solar altitude during the winter months are between 0 and 20, so exactly the range where the efficiency is underestimated. But still we get a higher power output than the reference model. That seems a bit odd... Any idea?

FabianHofmann:
Code-wise I have nothing to complain about :) Everything is super readable and well-structured

pz-max:
Hi Johannes, amazing contribution! :1st_place_medal: 
The Jupyter scripts looks also great. I like that you are verifying the model by comparing it to PS10 and also changing some of the datasets. Contentwise, everything seems alright. I would also forward this jupyter script as an initial draft to get some more qualified comments.

A little comment on the notebook:
As partly done, split cells in physics and plotting part i.e. cell 9, 16, 19, 22

euronion:
Thank you  both of you for your feedback!

@FabianHofmann 
>     1. I am wondering why the efficiency map is not defined for a azimuth range from 0 to 360. The ranges 0-70 (roughly) and 270 - 360 are missing. Does that mean a solar position in the north is not covered/supported?

Oops. You're absolutely right. Chile, South Africa, Australia, ... would all have suffered from this mistake. I corrected it now by mirroring the efficiency in the interval [90°,270°] to [90°,-90°=270°] with the mirror plane being 90°. Documentation is also updated. Thanks!

>     2. I am wondering whether the underestimated efficiencies for low solar altitudes play a role. These have a strong relative deviation from the SAM data. For example: For SAM the combination of altitude=8, azimuth=280 leads to en efficiency of ~10%. The same combination on the derived efficiency map gives  efficiency ~0%. Do you see any harmful potential there?

I don't think so, at least not in this example. A low solar altitude also coincides with a low direct solar irradiation. An solar altitude of e.g. 8° rarely happens when there is significant irradiation. Here's two figures showcasing this:

* In the top picture time vs. efficiency for the PS10 location in Spain during winter solistice. There are 4 points per day with significantly lower efficiency during sun rise and sun set.
* In the lower picture the DHI for the same location and time. Notice that for only 3 of the 4 time points with low efficiency the DHI is != 0. Simulatenously for those 3 points the irradiation is already low and the contribution to the daily output less significant anyway.

image

I agree that this is an effect which contributes a little bit. It is also something a user can improve for themselve by using a better efficiency map / CSP installation config as input. Lastly I think it is something that - as long as we are using hourly resolved data sets - we will always struggle with: The comparison of the output of a single plant at high temporal resolution is the edge case to which `atlite` can / should not be applied IMHO.

>     3. The underestimated efficiencies for low solar altitudes stand in contrast to the overestimated power outputs in winter. In the example the solar altitude during the winter months are between 0 and 20, so exactly the range where the efficiency is underestimated. But still we get a higher power output than the reference model. That seems a bit odd... Any idea?

**I presume this is an issue with the lack of atmospheric attenuation (For now we're using clearsky conditions).** This should lead to problems especially during the winter season in Spain. Inspecting the system in SAM yields the following picture for January:

image

In the top picture you see on the left y-axis in blue the output of the solar field to HTF (heat transfer fluid) in MW_th. In orange on the right the wind speed at that time. High wind speeds and low generation correlate (during daytime).
In the bottom picture for the same time range incident DNI (left axis, red) vs. relative humidity (black, right axis) on the solar field. Notice how they also negatively correlate.

Let's see what Max's expert say about this. :)

pz-max:
Hi Johannes,
Maybe worth adding at the bottom of the script some future work people can catch up on? > For now we're using clearsky conditions. There are ways to correct for the attenuation (based on humidity, altitude, dust). Maybe something for future work? BR, Max
________________________________

euronion:
@pz-max : will do. I'll add a note to the notebook and create+link an issue in the GH repository with "help wanted".

- [ ] Note to myself: Do the above

euronion:
I did a quick review on alternative datasets which might yield better results, especially for CSP data which relies on direct irradiation data on the surface, which in turn is influenced by the aerosol and cloud data used in each dataset. NREL's SAM uses the NSRDB which tries to be better in that regard than e.g. SARAH, which only uses a constant annual aerosol specification.
I created a separate issue #213 as a result of this investigation.

But more importantly:
During this review I noticed that the weather data set used by my SAM installation was off (filename reported a different location than the file contents listed). So the major deviations I saw in my comparisons and listed in the earlier version of the `examples/working-with-csp.ipynb` notebook were due to misaligned input data. (The data SAM downloaded was for somewhere in South Eastern Europe).

With the data for the correct location the deviations are much smaller and IMHO qualitatively totally acceptable, e.g.:

image

If the checks all pass I'd say we're ready to merge.

fneum:
Really nice :)

PR #195, opened 2021-11-17
Explicit units and meaning of returned DataArrays

euronion:
I'm not 100% satisfied with the PR, it feels a bit hacky. Nevertheless...

Explicitly provide units and a name for returned DataArrays created by `.wind(...)` and `.pv(...)` function.

## Description

For `.wind(...)` the location to set name and unit is straightforward.
For `.pv(...)` I'm not an expert for the Boilfinger model and couldn't find the correct location; I would assume the unit is identical to the Huld model, but since I am not 100% sure I decide to set unit and name inside the specific model.

Inside `.convert_and_aggregate(...)` where the DataArrays are manipulated the unit/name are either changed or keeping of the attributes is made explicit via `keep_attrs=True` (otherwise they would be implicitly removed by `xarray`).

Caveat: Setting of units relies on some conventions, e.g. for the units assumed in the panelconfig and turbineconfig files. However the code already makes these assumptions, so it is not a new hard-coded reliance that is introduced.

## Motivation and Context

Units until now were not made explicit for these cases and confusing.

## How Has This Been Tested?
Local testing.

## Type of change

## Checklist

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -125,7 +125,7 @@ def convert_and_aggregate(
             return maybe_progressbar(res, show_progress, **dask_kwargs)
         else:
             res = da.sum("time", keep_attrs=True)
-            return maybe_progressbar(res,  show_progress, **dask_kwargs)
+            return maybe_progressbar(res, show_progress, **dask_kwargs)
 
     if shapes is not None:
         geoseries_like = (pd.Series, gpd.GeoDataFrame, gpd.GeoSeries)
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -124,9 +124,8 @@ def convert_and_aggregate(
             res.attrs["units"] = "p.u."
             return maybe_progressbar(res, show_progress, **dask_kwargs)
         else:
-            return maybe_progressbar(
-                da.sum("time", keep_attrs=True), show_progress, **dask_kwargs
-            )
+            res = da.sum("time", keep_attrs=True)
+            return maybe_progressbar(res,  show_progress, **dask_kwargs)
 
     if shapes is not None:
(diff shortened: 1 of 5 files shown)


atlite/convert.py
```diff
@@ -120,9 +120,15 @@ def convert_and_aggregate(
                 "given for `per_unit` or `return_capacity`"
             )
         if capacity_factor:
-            return maybe_progressbar(da.mean("time"), show_progress, **dask_kwargs)
+            return maybe_progressbar(
+                da.mean("time").assign_attrs(units="p.u.").rename("capacity factor"),
+                show_progress,
+                **dask_kwargs,
+            )
```
FabianHofmann, line 127:
```suggestion
            res = da.mean("time").rename("capacity factor")
            res.attrs["units"] = "p.u."
            return maybe_progressbar(res, show_progress, **dask_kwargs)
```

atlite/convert.py
```diff
@@ -120,9 +120,15 @@ def convert_and_aggregate(
...
         else:
-            return maybe_progressbar(da.sum("time"), show_progress, **dask_kwargs)
+            return maybe_progressbar(
+                da.sum("time", keep_attrs=True), show_progress, **dask_kwargs
+            )
```
FabianHofmann, line 131:
```suggestion
            res = da.sum("time", keep_attrs=True)
            return maybe_progressbar(res,  show_progress, **dask_kwargs)
```

FabianHofmann (commented):
Great! Good that you have a look at that! Just a picky request on aligning the code style in the convert function to the other parts with returns

euronion:
Picky is welcome ;-)

PR #199, opened 2021-11-23
Calculate solar position already during cutout creation + add time-shift for ERA5

### LINKED ISSUE #158: [JOSS] Solar position calculation introduces time shift error for ERA5
I think, when simulating PV production, the solar position calculation is performed for the edge of each weather interval, introducing an effective time shift error when simulating solar output using interval-averaged weather like ERA5.

Note that I am new to ERA5 (and atlite), so it's possible that my understanding of its conventions is incorrect.

## Description
I don't have code to reproduce, but this is the reasoning (please check that my understanding is actually correct!):
- The relevant ERA5 weather parameters are interval averages (rather than instantaneous).  I think this is the case because `fdir` etc are included in Table 3 here, which lists the "accumulated" fields (as opposed to "instantaneous" fields): https://confluence.ecmwf.int/pages/viewpage.action?pageId=82870405#ERA5:datadocumentation-Table3
- The timestamp index of the xarray data set fetched from the CDS API represents the end of the interval, e.g. a `ds['time']` of `14:00:00` refers to the interval `13:00` to `14:00`.
- Using solar position for `14:00` to transpose the average irradiance from `13:00` to `14:00` introduces an effective time shift error in the transposed irradiance, meaning the simulated PV output will be shifted ~30 minutes in time.  See https://pvlib-python.readthedocs.io/en/latest/auto_examples/plot_interval_transposition_error.html

There appears to be a small timeshift in the plot of Cell 16 in the historic-comparison-germany notebook -- it may be from this issue, although I did not think through which direction the time shift error is and whether it matches the shift in that notebook.

## Expected Behavior
A better alternative is to calculate solar position in the middle of the interval instead of one edge.  It may also be desirable to handle the sunrise and sunset hours separately (see Section 4.3 here: https://www.nrel.gov/docs/fy15osti/64102.pdf) though that level of detail might be out of scope here.

## Actual Behavior
Sun position is calculated for the end of the interval, which is not as representative of the average position across the interval.

## Error Message
This is not a programming error, just a conceptual one. 

## Your Environment

* The `atlite` version used: 0.2.5.dev0+g97ff81c.d20210613
* How you installed `atlite` (`conda`, `pip` or `github`): github
* Operating System: Windows
* My environment: not relevant, but happy to provide if necessary

xref https://github.com/openjournals/joss-reviews/issues/3294

euronion:
Closes #158 .

* Solar position will be calculated during cutout creation
* `atlite.pv.solar_position.SolarPosition` results become part of the `"influx"` feature
* Solar position for ERA5 is calculated with an time-shift of "-30min" to take into account the fact that ERA5 variables are aggregated (right interval border, see #158 )
* `atlite.pv.solar_position.SolarPosition(...)` is backwards compatible (for now) and acts as a pure getter if the variables are present in the cutout, else it calculates them.

## How Has This Been Tested?

Tested locally on two cutouts (one ERA5, one SARAH)

image

## Type of change

## Checklist

- [ ] ~I have documented the effects of my code changes in the documentation `doc/`.~
- [ ] ~I have added newly introduced dependencies to `environment.yaml` file.~

## TODOs

- [x] Release notes
- [x] Suppress deprecation warning during cutout creation

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -14,10 +14,15 @@ Release Notes
 * Atlite now supports calculating dynamic line ratings based on the IEEE-738 standard (https://github.com/PyPSA/atlite/pull/189).
 * The wind feature provided by ERA5 now also calculates the wind angle `wnd_azimuth` in range [0 - 2π) spanning the cirlce from north in clock-wise direction (0 is north, π/2 is east, -π is south, 3π/2 is west).
 * A new intersection matrix function was added, which works similarly to incidence matrix but has boolean values.
-
+* The solar position (azimuth and altitude) are now part of the cutout feature `influx`. Cutouts created with earlier versions will become incompatible with the next major.
 
 * Automated upload of code coverage reports via Codecov.
 * DataArrays returned by `.pv(...)` and `.wind(...)` now have a clearer name and 'units' attribute.
+
+**Bug fixes**
+* The solar position for ERA5 cutouts is now calculated for half a time step earlier (time-shift by `cutout.dt/2`) to account for the aggregated nature of
+  ERA5 variables (see https://github.com/PyPSA/atlite/issues/158). The fix is only applied to newly created cutouts. Previously created cutouts do not profit
+  from this fix and need to be recreated `cutout.prepare(overwrite=True)`.
 
 Version 0.2.5 
 ==============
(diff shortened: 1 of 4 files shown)


FabianHofmann (approved):
Great @euronion ! I think that is a good solution. We should open an issue to address the deprecation for the release after the next one.

euronion:
I've created the issues and tagged it with `v0.4` (which should be the next-next major version, right?).

Also I noticed a small bug I introduced: The previous implementation assumed hourly time-steps for ERA5, it now supports arbitrary time-steps for the `time_shift` with ERA5 cutouts.

I'll merge as soon as CI clears the PR.

PR #202, opened 2021-12-03
update xarray version

FabianHofmann:
chunksizes (used by new Solarposition function) requires xarray >= v0.20

## Type of change

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ setup(
         "pandas>=0.25",
         "bottleneck",
         "numexpr",
-        "xarray>=0.16.2",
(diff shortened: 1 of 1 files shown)


euronion:
Thanks, that was unintentional. The original version of the `SolarPosition` function (which has been part of `atlite` for long) used `.chunks` instead of `.chunksizes`. Both are synonym. If we don't want to increase the min `xarray` version we could also revert to `.chunks` instead.

FabianHofmann:
I think it's fine to catch up with the up-to-date version

PR #203, opened 2021-12-03
Better handling of stacked spatial dimension in `matrix` argument 

FabianHofmann:
There is a nasty misalignment of the matrix coordinates with the cutout coordinates in the land-use example. To ensure compatibility of matrices and cutout in the future, this PR proposes some changes to the convert_and_aggregate function. 

# Changes

* Check coordinate alignment of a `DataArray`-matrix with `cutout.grid` in `convert_and_aggregate`.
* `convert_and_aggregate` now supports a matrix of shape (N, X, Y). In this case the matrix must be a `xarray.DataArray` with dimensions `x` and `y` equal to those of `cutout.data`. 
* Due to ambiguity, conversion functions (`.pv(...)`, `.wind(...)` etc.) now raise an `ValueError` if shapes and matrix are given. 

## Type of change

## Checklist

- [ ] ~I have added newly introduced dependencies to `environment.yaml` file.~


FabianHofmann:
I would merge this if no one has an opposing opinion @euronion @fneum ?

fneum:
No opposition from my side!

atlite/convert.py
```diff
@@ -789,7 +824,7 @@ def line_rating(cutout, shapes, line_resistance, **params):
     I = cutout.intersectionmatrix(shapes)
     rows, cols = I.nonzero()
 
-    data = cutout.data.stack(spatial=["x", "y"])
+    data = cutout.data.stack(spatial=["y", "x"])
```
euronion, line 827: Based on which convention are we stacking?

atlite/convert.py
```diff
@@ -62,10 +62,11 @@ def convert_and_aggregate(
 
     Parameters
     -----------
-    matrix : N x S - xr.DataArray or sp.sparse.csr_matrix or None
+    matrix : N x S or N x X x Y- xr.DataArray or sp.sparse.csr_matrix or None
```
euronion, line 65: TBH I don't like it.

`matrix` by itself is already a slightly confusing construct. Overloading it now to be able to accept a different datatype with a different dimension doesn't help with the confusion, but increases it.

Can we have a dedicated argument for `xr.DataArrays`? (I like the idea but not the implementation as an overloaded function parameter)

atlite/convert.py
```diff
@@ -128,22 +129,56 @@ def convert_and_aggregate(
...
+
+        if shapes is not None:
+            raise ValueError(
+                "Passing matrix and shapes is ambiguous. Pass " "only one of them."
+            )
+
+        if isinstance(matrix, xr.DataArray):
+
+            if {"x", "y"}.issubset(matrix.dims):
+                matrix = matrix.stack(spatial=["y", "x"])
```
euronion, line 142: Based on which convention are we stacking here? If this is following a convention already followed somewhere else in `atlite`, maybe we can not hardcode it but reference the appropriate object and follow its stacking rule?

atlite/convert.py
```diff
@@ -128,22 +129,56 @@ def convert_and_aggregate(
...
+                raise ValueError(
+                    "Matrix spatial coordinates not aligned with cutout spatial "
+                    "coordinates."
+                )
+
+            if index is None:
+                index = matrix
+
+        if not matrix.ndim == 2:
+            raise ValueError("Matrix not 2-dimensional.")
```
euronion, line 156: Just to emphasize my point from above: This error message is very confusing to people leveraging `matrix` with a 3D `xr.DataArray`.

euronion:
@FabianHofmann Sorry, I'm not happy with it. See comments.

FabianHofmann:
@euronion thanks for your thoughts! Two things 

1. You are right about the passing a non-stacked 3-dimensional DataArray as a matrix. This bloats the set of possible arguments. 
2.  However passing a `DataArray` with stacked coordinates was supported all the time and I strongly advocate to support it in the future. Two weeks ago I was "debugging" with an atlite user for 1 and a half hours, before recognizing that his matrix was stacked in the wrong way. The atlite conversion for a stacked coordinate axis is that both `x` and `y` are increasing, but `x` rolls over all values for each value in `y`, so `y` is monotonically increasing and `x` not (just like in `cutout.grid`). This is achieved by `ds.stack(spatial=['y', 'x'])`. When `matrix` is a `DataArray`, the alignment of the stacked coordinates with this convention was not checked before. Now the alignment is double-checked before converting it to a sparse matrix. The example and the dynamic line rating code had a wrong stacking, this was also addressed.

euronion:
>     1. You are right about the passing a non-stacked 3-dimensional DataArray as a matrix. This bloats the set of possible arguments.

I do like the idea of being able to pass an unstacked 3D DataArray. To me passing a 3D array feels more natural. But maybe not by overloading the same argument, but with a new argument, e.g. `matrix_3d`?

>     2. However passing a `DataArray` with stacked coordinates was supported all the time and I strongly advocate to support it in the future.

Of course, keep the backwards compatability! Sorry, I missed that when looking at the code. 

> This is achieved by `ds.stack(spatial=['y', 'x'])`.

I get your point. It's just that I don't like hardcoding things like 'y','x' and 'spatial' into the code, because that's what causes misalignment if the hardcoding is in multiple locations.
You can access the order of stacked dimensions via
```
da.coords.indexes['spatial'].names        # FrozenList(['y', 'x'])
```

FabianHofmann:
Alright, I think we can pull this in. If we want to support a 3d matrix as a separate argument, we can tackle that another time. Any objection @euronion?

euronion:
LGTM

PR #205, opened 2021-12-09
WIP: Cutout - Access to irradiation data

Tasqu:

1. Added a new optional keyword `irradiance='total'` to `TiltedIrradiance()` to determine whether to return *total*, *direct*, *diffuse*, or *ground* irradiation. Default assumption is *total* to preserve the original functionality.
2. Added `convert_irradiation()` and `irradiation()` methods to the `Cutout` class to access the raw irradiance data calculated via `TiltedIrradiance()`.
3. Also added a new example for demonstrating a use case for the feature.

### pv/irratiation.py `TiltedIrradiation()`

Added a new optional keyword `irradiation="total"` to define what irradiation quantity should be returned. Previously, `TiltedIrradiation()` only returned the total irradiation, but for my purposes, I needed access to *direct* and *diffuse* irradiation on a tilted surface separately. The method defaults to `irradiation="total"` in order to minimise the impact of this change on the rest of `atlite`.

### convert.py `convert_irradiation()` and `irradiation()`

Added two new methods for accessing the raw irradiation data, `convert_irradiation()` and `irradiation()`. These are basically just reduced versions of the `convert_pv()` and `pv()` methods, where we skip the final conversion from solar irradiation to PV generation.

### cutout.py `irradiation`

Added the `irradiation()` method to the `Cutout` class in order to access the raw irradiation data.

### examples/finnish_building_stock_weather_aggregation.ipynb

Added a new example to demonstrate using the added features.

### examples/finnish_building_stock_weather_aggregation.ipynb.license

License file for the new example, `GPL-3.0-or-later`

### doc/examples/finnish_building_stock_weather_aggregation.nblink

`nblink` file for the new example.

WIP

## Motivation and Context

I'm working on modelling the heat demand of the Finnish building stock, its potential flexibility, as well as its value in large-scale energy system context. For this reason, I need to be able to calculate historical weather time series over large geographical areas, weighted based on where the majority of the heated volume of the building stock is located. We've used in-house Python codes for these kinds of things previously, but those have fell into disrepair after the original developers have moved on.

Seeing that `atlite` was already capable of doing all the calculations I need, I decided it would easier to tweak `atlite`, than to try and learn and repair our old in-house codes. However, the current `atlite` `master` doesn't actually contain methods for accessing the tilted solar irradiation directly, as it is only used for PV and solar thermal conversion. Thus, I added the necessary methods myself.

## How Has This Been Tested?

So far, mostly just using the examples. The new one works as intended, and I don't think I broke any of the old ones.

WIP

## Type of change

## Checklist


euronion:
Hi @Tasqu !

Thanks for your PR. I'll have a look at it later this week or next week.

atlite/convert.py
```diff
@@ -493,6 +493,76 @@ def wind(cutout, turbine, smooth=False, **params):
...
+        'atlite.pv.orientation.make_*' functions.
+    irradiation : str
+        The irradiation quantity to be returned. Defaults to "total" for total
+        irradiation. Other options include "direct" for direct irradiation,
+        "diffuse" for diffuse irradation, and "ground" for ground irradiation.
+    clearsky_model : str or None
+        Either the 'simple' or the 'enhanced' Reindl clearsky
+        model. The default choice of None will choose dependending on
+        data availability, since the 'enhanced' model also
+        incorporates ambient air temperature and relative humidity.
```
euronion, line 536: I know that this is what the `.pv(..)` documentation says; which by itself isn't very comprehensive:

The `clearsky_model` is only applied for determining shares of direct and diffuse irradiation *if* they are not already provided by the `cutout` data. I.e. if you use `ERA5` or `SARAH` data, the `clearsky_model` parameter doesn't do anything and no realsky model is applied. (Because cutouts from both datasets include `influx_direct` and `influx_diffuse` variables and thus no calculation happens in `TiltedIrradiation(...)`). This is part of a larger portion of legacy code which was required in the past when we were supporting datasets with only `influx` variables.

Maybe that's also relevant for your implementation?

atlite/pv/irradiation.py
```diff
@@ -157,6 +157,7 @@ def TiltedIrradiation(
     trigon_model,
     clearsky_model,
     altitude_threshold=1.0,
+    irradiation="total",
```
euronion, line 160: This might cause confusion, as `total` could also refer to `direct + diffuse + ground reflected + lost`. E.g. in the case of incident irradiation `total` is `direct+diffuse`.

Two suggestion:
1. Call it `combined` rather than `total`.
2. Add a docstring to `TiltedIrradiation` based on your docstring for the `irradiation(...)` method
```
    irradiation : str
        The irradiation quantity to be returned. Defaults to "total" for total
        irradiation. Other options include "direct" for direct irradiation,
        "diffuse" for diffuse irradation, and "ground" for ground irradiation.
```
where you explain what `total` (or `combined`) refers to.

atlite/pv/irradiation.py
```diff
@@ -186,6 +187,7 @@ def clip(influx, influx_max):
         diffuse_t = (1.0 + cos_surface_slope) / 2.0 * diffuse + _albedo(
             ds, influx
         ) * influx * ((1.0 - cos_surface_slope) / 2.0)
+        ground_t = None
```
euronion, line 190: I believe this is wrong.

1. This will probably lead to errors with the `.convert_and_aggregate(...)` method, as the type is not an `xr.DataArray`.
2. The ground reflected irradiation is contained in the `diffuse_t` calculation above; c.f. with second term containing `_albedo`.

--> separate the calculation of `diffuse_t` and `ground_t` for `trigon_model == "simple"` and return the correct `ground_t`.

atlite/pv/irradiation.py
```diff
@@ -199,11 +201,20 @@ def clip(influx, influx_max):
...
         total_t = (direct_t + diffuse_t + ground_t).rename("total tilted")
 
+    if irradiation == "total":
+        result = total_t
+    elif irradiation == "direct":
+        result = direct_t
+    elif irradiation == "diffuse":
+        result = diffuse_t
+    elif irradiation == "ground":
+        result = ground_t
```
euronion, line 211: The returned `xarray` objects should be named correctly, i.e. add
```
results = results.rename("total tilted")
```
and the respective other names.

We're also trying to consistently add `units` to the `xarray` objects whenever we touch them (in the past we did not provide units, which can be confusing). Can you add the appropriate unit to `results` as well?
Like e.g. here:
https://github.com/PyPSA/atlite/blob/40dd1fcc8f494379acd9f837e7ef1e0964cae3cd/atlite/pv/solar_position.py#L131

euronion (changes requested):
Thanks for your PR.

Sorry, it took a while for me to sit down and think it through. I have a few comments, some concerning documentation (for others and our future selves), others involve functionality.

Also two remarks on the notebook:

* Can you think of a shorter title for the notebook example? Long titles render badly in the table of contents in the [documentation](https://atlite.readthedocs.io/en/latest/examples/create_cutout_SARAH.html).
* The notebook is not executed, but uploaded without cell output. We try to include meaningful cell output with the examples (figures or display variable values). It'd be great if you could include meaningful cell output with the example, see e.g. [this example](https://atlite.readthedocs.io/en/latest/examples/historic-comparison-germany.html).

Let me know where you (dis-)agree with my comments.

Tasqu:
@euronion I've been a bit busy lately, so I don't know when I'll have the time to address your comments above. Hopefully within a few weeks, but I can't make any promises...

Still, thanks for going through my pull request, and for the thorough feedback!

euronion:
Great. 👍 No need for promises, As the old saying goes: “Your contributions are never late, nor are they early, they arrive precisely when you meant them to."

atlite/convert.py
```diff
@@ -493,6 +493,76 @@ def wind(cutout, turbine, smooth=False, **params):
     )
 
 
+# irradiation
+def convert_irradiation(
+    ds, orientation, irradiation="total", trigon_model="simple", clearsky_model="simple"
```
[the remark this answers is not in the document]
Tasqu: I think I just did exactly what the `pv` and `convert_pv` methods do, since I wasn't sure how you wanted defaults like this to be handled. Is there something I'm missing, or does the `convert_pv` method also have a redundant default `clearsky_model="simple"`?

atlite/convert.py
```diff
@@ -493,6 +493,76 @@ def wind(cutout, turbine, smooth=False, **params):
...
+):
+    """
+    Calculate the total, direct, diffuse, or ground irradiation on a tilted
+    surface.
+
+    Parameters
+    ----------
+    orientation : str, dict or callback
+        Surface orientation can be chosen from either
+        'latitude_optimal', a constant orientation {'slope': 0.0,
```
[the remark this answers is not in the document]
Tasqu: Frankly, using `latitude_optimal` doesn't really make much sense to me either. I suppose currently it results in angling the surface so that it collects most of the irradiation over a year, but I'm not sure what it would be used for. My use cases all define a constant known orientation, and I just used the same orientation functions as the PV conversion functions for convenience. I suppose I could clarify the docstring that it shouldn't probably be used here.

atlite/convert.py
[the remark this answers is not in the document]
Tasqu: Ok, regarding your later comment of how the `clearsky_model` is used, I think it's indeed better to leave the default value `None` for both functions.

atlite/convert.py
```diff
@@ -493,6 +493,76 @@ def wind(cutout, turbine, smooth=False, **params):
...
+        'azimuth': 0.0} or a callback function with the same signature
+        as the callbacks generated by the
+        'atlite.pv.orientation.make_*' functions.
+    irradiation : str
+        The irradiation quantity to be returned. Defaults to "total" for total
+        irradiation. Other options include "direct" for direct irradiation,
+        "diffuse" for diffuse irradation, and "ground" for ground irradiation.
```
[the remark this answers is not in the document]
Tasqu: Good point, I clarified the docstrings

atlite/convert.py
Tasqu: Shouldn't be relevant as far as I can tell. My implementation basically just needs access to the raw irradiation data already used by the PV functions, but without the radiation-to-generation conversion.

Tasqu: Added a clarification in the docsrting.

atlite/pv/irradiation.py
Tasqu: I think I'll stick with `total` for now, as it seems to me the resulting `xarray.Dataset` is currently renamed to `total tilted` anyway. I think renaming to `combined` would result in undue confusion. I tried adding a docstring though, but I must admit I'm not 100% clear on whether I got the explanations for all the input parameters correct.

atlite/pv/irradiation.py
Tasqu: Ground reflected irradiation now separated from `diffuse_t` for the `simple` `trigon_model`. I hope I did it right, as I'm by no means an expert in solar irradiation calculations.

atlite/pv/irradiation.py
Tasqu: Renamed and unit added to the `result`.

Tasqu:
@euronion Ok, I think I've addressed all of your previous comments. I also revised the example so that it doesn't display any unnecessary intermediate output, and included the outputs of interest into the notebook.

Tasqu (commented):
Review done?

FabianHofmann (changes requested):
Hey @Tasqu, this is really impressive work, in particular the example. Sorry for taking so long, I'm regularly finding myself postponing reviews of larger PRs. 

One point about the example notebook. The projection of the population layout is a bit bloated and probably not fully correct. For the re-projection, could you try something like this:

```python 
finland_population_density.rio.reproject(cutout.crs, shape=cutout.shape, transform=cutout.transform)
```

After that is fixed, I am happy to merge :)

Tasqu:
Ok, I revised the example to be a bit more dense *(and fixed a few typos)*. I also went through the pull request checklist again, but it seems that `pytest` doesn't pass at the moment *(seems like `master` is suffering from the same issue?)*

Let me know if there's still some stuff you need improved.

FabianHofmann (approved):

FabianHofmann:
@euronion you have to approve that you are happy with your requested changes now

euronion (approved):
Thank you for the changes!

FabianHofmann:
@Tasqu would you be fine if we squash this before merging?

Tasqu:
@FabianHofmann I have no objections, it's your codebase after all 😉

PR #215, opened 2022-01-28
correct equation natural convection

p-glaum:
There is a small mistake in the natural convection equation. It is supposed to be the square root of rho. 
image


FabianHofmann (approved):
@p-glaum great that you are double-checking rigorously! (I think this hardly changes our results, as it counts for zero wind speed).

PR #217, opened 2022-02-02
Revise historic comparison example

FabianHofmann:
As mentioned in the pypsa mailing list, the azimuth in the example was mistakenly set to 0, is it corrected to 180 now.  

Also the usage of `Cutout.layout_from_capacity_list` was imposed (prevents warnings, is leaner)

EDIT: This PR supersedes and includes the proposed changes in #218 due to direct effect on the NB (and these should not be updated too often due to their large file size). Taken from #218:

The current implementation does not convert degrees to radian in the `make_latitude` and `make_latitude_optimal` function. This fixes it. 

## Type of change

## Checklist


euronion (changes requested):
> Whereas in the spatial aggregation the optimal slope performs better, on a detailed level the spread is very high. Thus the steady slope of 30° might be a saver choice knowing that this tends to underestimate.

This sentence no longer makes sense (above cell 18).

Maybe add a note on the orientation choosen in the example?
```
orientation={'slope': 30., 'azimuth': 180.}
```

(e.g. south orientation ( `'azimuth': 180.` ) and prominent slope 30-35° for PV in Germany (`'slope':30.` ).

euronion:
Great, LGTM

Issue #211, opened 2022-01-13
Split influx from solar_position feature


FabianHofmann:
Due to the currently overloaded `influx` feature, we should split group the `solar_altitude` and `solar_azitmuth` into a new feature `solar_position`.

PR #222, opened 2022-02-23
purge influx feature 

### LINKED ISSUE #212: Remove atmospheric_insolation from SolarPosition function
This variable is neither needed with the current modules nor at the right place. If needed its calculation should be part of the preparation process

FabianHofmann:
Closes #212 
Touches #211 

Initial attempt was to solve #211, but is turns out to be tricky. The problem is that the era5 module relies on get_<feature> functions which take the era5 retrieval parameters as an argument. The SolarPosition function however requires a dataset with the complete set of coordinates. It would be necessary to pass the `cutout.data` as an argument to the `get_solar_position` function. This requires changes one have to think through a bit. 
The same counts for the sarah module. Its `get_data` function does not use the `feature` argument so far. It would be necessary to make separate `get_influx` and `get_solar_position` functions. I would postpone that change to another time. 

In order to make the `influx` feature still a bit handier, I removed the solar athmospheric insolation variable from the `influx` feature as well as from the `SolarPosition` function. The variables were renamed to `solar_azimuth` and `solar_altitude`. 

## Motivation and Context

Make the `influx` feature leaner. Make data variables accessible via `cutout.data.solar_altitude`/ `cutout.data.solar_azimuth`

## How Has This Been Tested?

No test added.

## Type of change

## Checklist

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -22,6 +22,7 @@ Release Notes
 * Due to ambiguity, conversion functions (`.pv(...)`, `.wind(...)` etc.) now raise an `ValueError` if shapes and matrix are given. 
 * Atlite now supports calculating of heat pump coefficients of performance (https://github.com/PyPSA/atlite/pull/145).
 * Enabled the GitHub feature "Cite this repository" to generate a BibTeX file (Added a `CITATION.cff` file to the repository).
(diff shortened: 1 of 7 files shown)


euronion (commented):
I didn't test it, in general it LGTM. Thanks! :)

PR #233, opened 2022-04-11
Add NREL turbine data to atlite

thesethtruth:
## Description
Adds more turbine .yaml to the database.

## Motivation and Context
Some turbines in the current database are a bit outdated. Moreover, if used in ESOM, users might want to use power curves for future / expected turbine types (with a higher rated power and different power curve). This is done based on the NREL turbine data set, which I've parsed to the atlite format using some basic python and Jinja2 templating (see: [turbine parsing repo](https://github.com/thesethtruth/NREL-to-atlite-turbines)). 

In our current project we use PyPSA-eur and felt like this feature was limiting. This was discussed with @fneum.

## How Has This Been Tested?
Not yet.

## Type of change
Extension of existing data

Feel free to propose updates / changes. Let me know what you need to use this data.


atlite/resources/windturbine/2016CACost_NREL_Reference_10MW_205_offshore.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# Turbine characteristic data
+name: 2016CACost_NREL_Reference_10MW_205
+# The manufacturer of this turbine
+manufacturer: NREL
+# Link to the original datasheet of the turbine
+source: https://nrel.github.io/turbine-models/2016CACost_NREL_Reference_10MW_205.html
+
+# Hub height in meters
+HUB_HEIGHT: 205.0
```
euronion, line 13: `HUB_HEIGHT` seems incorrect for most (all?) turbines. For this turbine it would be `125 m` according to the linked source. The `205 m` are the rotor diameter.

atlite/resources/windturbine/2016CACost_NREL_Reference_10MW_205_offshore.yaml
```diff
@@ -0,0 +1,19 @@
+# SPDX-FileCopyrightText: 2022 The Atlite Authors
+#
+# SPDX-License-Identifier: CC-BY-4.0
+
+# Turbine characteristic data
+name: 2016CACost_NREL_Reference_10MW_205
```
euronion, line 6: We should use a consistent file naming scheme for the files and turbines:

Until now we are using

`<manufacturer>_<manufacturer turbine name>_<power in MW>.yaml`

Could you adjust the file names accordingly? For the NREL turbines it might make sense to e.g. have `NREL_ReferenceTurbine_ATB2020_10MW.yaml`

The `name: ` value inside the `.yaml` files can stay like the original used in NREL's GitHub repo.

euronion (changes requested):
Hi @thesethtruth ,

thanks for the PR! You are right: We have been putting off introducing new turbines for a long time. There are a two things which need to be adressed. See the comments in one of the files (the comments apply to all files).

Let me know if you have any questions.

atlite/resources/windturbine/2016CACost_NREL_Reference_10MW_205_offshore.yaml
thesethtruth: Yes, you are absolutely right! I assumed the rotor diameter as the hub height from the file name, now I've sourced it from the .rst in their docs.

thesethtruth:
Comments from @euronion processed. 

If in the future you'd want to parse the data yourself; I've automated most in https://github.com/thesethtruth/NREL-to-atlite-turbines. 

Cheers!

fneum:
GTM! Thanks!

euronion:
Ah; nearly missed it: Can add a oneliner to the RELEASE_NOTES mentioning new turbines ? @thesethtruth

thesethtruth:
Done! See below:

* Atlite now includes the reference turbines from the NREL turbine archive (see: https://nrel.github.io/turbine-models/). Available turbines can be consulted using `atlite.windturbines` and can be passed as string argument, e.g. `coutout.wind(turbine)`.

euronion:
Great, thanks!

coroa:
Thanks for that merge, quite helpful!

As a further idea you could consider adding your existing parsing code directly to atlite as an extension to `get_windturbineconfig` in the same manner as we do for the oedb; so that one could use:
```python
cutout.wind(turbine="nrel:2016CACost_NREL_Reference_8MW_180")
```
to transparently fetch the turbine defined at https://nrel.github.io/turbine-models/2016CACost_NREL_Reference_8MW_180.html 

That would work by adding a new hook in `get_windturbineconfig`. https://github.com/PyPSA/atlite/blob/master/atlite/resource.py#L52-L53

And putting your code into a new function like `get_oedb_windturbineconfig` https://github.com/PyPSA/atlite/blob/2ab6c9e56db539ee3b6d5dfa9e12c1185b5954f9/atlite/resource.py#L238-L371

Maybe that is good first issue. But, many thanks either way!

PR #250, opened 2022-08-28
Issue227 - Turbine, Panel and CSP Installation Configs from local path

### LINKED ISSUE #227: Load wind turbine config from local path

LukasFrankenQ:
Closes #227.

Methods ``get_windturbineconfig``, ``get_solarpanelconfig`` and ``get_cspinstallationconfig`` load yaml files based on the passed
configuration name. Now, additionally, the methods recognize if a path to a locally stored config is passed.
The file is then loaded instead of a predefined config.

## Description

The methods run ``Path([the local path]).exists()`` to check if the path exists. If not, the methods assume that the user passed 
the name of a config, and proceeds as before the change.

## Motivation and Context

This is a small change, proposed a while ago by @fneum. I seems like a sensible addition.

## How Has This Been Tested?

I checked all three functions to make sure that both old and new behaviour are behaving as expected.

## Type of change

## Checklist

- [ ] I locally ran `pytest` inside the repository and no unexpected problems came up. (one apparently unrelated error)

### DIFF
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -53,10 +53,17 @@ def get_windturbineconfig(turbine):
         return get_oedb_windturbineconfig(turbine[len("oedb:") :])
 
     if isinstance(turbine, str):
-        if not turbine.endswith(".yaml"):
-            turbine += ".yaml"
+        
+        if not Path(turbine).exists():
 
-        turbine = WINDTURBINE_DIRECTORY / turbine
+            if not turbine.endswith(".yaml"):
+                turbine += ".yaml"
+
+            turbine = WINDTURBINE_DIRECTORY / turbine
+        
+        else:
+
+            turbine = Path(turbine)
 
     with open(turbine, "r") as f:
         conf = yaml.safe_load(f)
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -106,19 +106,16 @@ def get_cspinstallationconfig(installation):
     config : dict
         Config with details on the CSP installation.
     """
-
         
     if not Path(installation).exists():        
 
         # if isinstance(installation, str):    # not sure what this does
         if not installation.endswith(".yaml"):
             installation += ".yaml"
-
         installation = CSPINSTALLATION_DIRECTORY / installation
     
     else:
         installation = Path(installation)
-    
 
     # Load and set expected index columns
     with open(installation, "r") as f:
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -42,24 +42,24 @@ def get_windturbineconfig(turbine):
 
     Parameters
     ----------
-    turbine : str
+    turbine : str or pathlib.Path
         Name of the local turbine file.
-        Alternatively a dict for selecting a turbine from the Open Energy
+        Alternatively, a dict for selecting a turbine from the Open Energy
         Database, in this case the key 'source' should be contained. For all
         other key arguments to retrieve the matching turbine, see
         atlite.resource.download_windturbineconfig() for details.
     """
 
+    assert isinstance(turbine, (str, Path))
+
     if isinstance(turbine, str) and turbine.startswith("oedb:"):
         return get_oedb_windturbineconfig(turbine[len("oedb:") :])
 
     if isinstance(turbine, str):
-
-        if not Path(turbine).exists():
-            if not turbine.endswith(".yaml"):
-                turbine += ".yaml"
-
-            turbine = WINDTURBINE_DIRECTORY / turbine
+        try:
+            turbine = windturbines[turbine.replace(".yaml", "").replace(".yml", "")]
+        except KeyError:
+            pass
 
     with open(turbine, "r") as f:
         conf = yaml.safe_load(f)
@@ -75,15 +75,13 @@ def get_windturbineconfig(turbine):
 def get_solarpanelconfig(panel):
     """Load the 'panel'.yaml file from local disk and provide a solar panel dict."""
 
+    assert isinstance(panel, (str, Path))
+
     if isinstance(panel, str):
-
-        if not Path(panel).exists():
-            if not panel.endswith(".yaml"):
-                panel += ".yaml"
-            panel = SOLARPANEL_DIRECTORY / panel
-
-        else:
-            panel = Path(panel)
+        try:
+            panel = solarpanels[panel.replace(".yaml", "").replace(".yml", "")]
+        except KeyError:
+            pass
 
     with open(panel, "r") as f:
         conf = yaml.safe_load(f)
@@ -96,7 +94,7 @@ def get_cspinstallationconfig(installation):
 
     Parameters
     ----------
-    installation : str
+    installation : str or pathlib.Path
         Name of CSP installation kind. Can either correspond to name of one of the files
         in resources/cspinstallation or be a local file.
 
@@ -106,15 +104,16 @@ def get_cspinstallationconfig(installation):
         Config with details on the CSP installation.
     """
 
-    if not Path(installation).exists():
+    assert isinstance(installation, (str, Path))
 
-        # if isinstance(installation, str):    # not sure what this does
-        if not installation.endswith(".yaml"):
-            installation += ".yaml"
-        installation = CSPINSTALLATION_DIRECTORY / installation
-
-    else:
-        installation = Path(installation)
+    if isinstance(installation, str):
+        try:
+            installation = cspinstallations[
+                installation.replace(".yml", "").replace(".yaml", "")
+            ]
+            installation = CSPINSTALLATION_DIRECTORY / installation
+        except KeyError:
+            pass
 
     # Load and set expected index columns
     with open(installation, "r") as f:
--- a/atlite/resource.py
+++ b/atlite/resource.py
@@ -59,9 +59,8 @@ def get_windturbineconfig(turbine):
                 turbine += ".yaml"
 
             turbine = WINDTURBINE_DIRECTORY / turbine
+
         
-        else:
-            turbine = Path(turbine)
 
     with open(turbine, "r") as f:
         conf = yaml.safe_load(f)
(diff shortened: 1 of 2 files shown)


atlite/resource.py
```diff
@@ -53,10 +54,12 @@ def get_windturbineconfig(turbine):
...
     if isinstance(turbine, str):
-        if not turbine.endswith(".yaml"):
-            turbine += ".yaml"
 
-        turbine = WINDTURBINE_DIRECTORY / turbine
+        if not Path(turbine).exists():
+            if not turbine.endswith(".yaml"):
+                turbine += ".yaml"
+
+            turbine = WINDTURBINE_DIRECTORY / turbine
```
euronion, line 62: Rather than presuming that if the Path does not exist that then it might be a pre-shipped one, we can check if it is a shipped config by checking:

```
turbine in windturbines
```

and then instead of manually constructing the path

```
turbine = windturbines[turbine]
```

as we have the turbines which are pre-shipped as well as their paths already read in `windturbineconfig`.

atlite/resource.py
```diff
@@ -34,10 +34,11 @@
 def get_windturbineconfig(turbine):
     """Load the wind 'turbine' configuration.
 
-    The configuration can either be one from local storage, then 'turbine' is
+    The configuration can be one from local storage, then 'turbine' is
     considered part of the file base name '<turbine>.yaml' in config.windturbine_dir.
-    Alternatively the configuration can be downloaded from the Open Energy Database (OEDB),
+    Alternatively, the configuration can be downloaded from the Open Energy Database (OEDB),
     in which case 'turbine' is a dictionary used for selecting a turbine from the database.
+    Finally, turbine can also be a path to a local config file.
```
euronion, line 41: We need to do some adjustements here, as `config.windturbine_dir` does not exist (very old legacy piece of doc which shouldn't be here).

atlite/resource.py
```diff
@@ -53,10 +54,12 @@ def get_windturbineconfig(turbine):
         return get_oedb_windturbineconfig(turbine[len("oedb:") :])
 
     if isinstance(turbine, str):
```
euronion, line 56: We could enforce the user to provide a `Path` object if the configuration is supposed to be read from a local path (instead of asking for a `str`. At the moment, if a `Path` object is provided the method fails.

That would also allow for better overloading of the argument :)

```suggestion
    if isinstance(turbine, Path):
```

atlite/resource.py
```diff
@@ -73,10 +76,14 @@ def get_solarpanelconfig(panel):
...
-            panel += ".yaml"
 
-        panel = SOLARPANEL_DIRECTORY / panel
+        if not Path(panel).exists():
+            if not panel.endswith(".yaml"):
+                panel += ".yaml"
+            panel = SOLARPANEL_DIRECTORY / panel
+
+        else:
+            panel = Path(panel)
```
euronion, line 86: See above for windturbine configs

atlite/resource.py
```diff
@@ -90,20 +97,24 @@ def get_cspinstallationconfig(installation):
...
+    if not Path(installation).exists():
+
+        # if isinstance(installation, str):    # not sure what this does
         if not installation.endswith(".yaml"):
             installation += ".yaml"
+        installation = CSPINSTALLATION_DIRECTORY / installation
 
-    installation = CSPINSTALLATION_DIRECTORY / installation
+    else:
+        installation = Path(installation)
```
euronion, line 117: See above for windturbine configs

atlite/resource.py
euronion: E.g.

The configuration can be either one of the configurations shipped with atlite, one loaded from a local path or retrieved from the Open Energy Database (OEDB):
If one of the names of the preshipped turbines from `atlite.resources.windturbines` is provided, that turbine configuration is loaded and returned.
If a `pathlib.Path` object is provided, this local path is read and the turbine configuration from this Path is parsed.
If a string starting with "oedb:<turbine name>" is provided, atlite will try to read the turbine configuration for this <turbine name> from OEDB.

euronion (changes requested):
Some ideas for improvements. What do you think?

NB: We need to be careful if we change how the of `get_windturbineconfig` is handled. If we only allow `str` values from `atlite.resource.windturbines` then that would be a breaking change which we should avoid.

LukasFrankenQ:
Hey guys, thank you for the suggestions (and the suggestion by @euronion on Discord). I believe that the following strikes a balance between both suggestions made, i.e. retains flexibility wrt being agnostic in Path vs str, being short, being robust against obscure errors like passing 'turbine_name.yml' instead of 'turbine_name.yaml'.
The panel example would go

`
if isinstance(panel, str):
    try:
         panel = solarpanel[panel.replace('.yaml', '').replace('.yml', '')]
    except KeyError:
         pass

with open(panel, 'r') as f:
    conf = yaml.safe_load(f)

return conf
`

The only problem with this is that if I had 'nonexistent_turbine.yaml', I will get a FileNotFoundError which is harder to interpret than a KeyError, but I dont see a way around this given that we allow str objects to be both a turbine name and a path.
And then I would of course update the doc, what do you think?

euronion:
I would prefer strongly enforcing `Path` for paths and only allowing `str` for turbine (and other) names @FabianHofmann .

We don't do it for cutouts partially due to legacy reasons and partially because we have the freedom to do so, as there are no "standard" cutouts which ship with `atlite`.

Standard use case in the past for the turbine name (and other configs) was to provide the name of the turbine which was translated into a path. Now we would just enforce this behaviour.

We could even think about omitting the `.replace('.yaml', '').replace('.yml', '')` from Lukas' proposal to make this even more clear: Either provide the name of the turbine or provide a path if you want to load a specific file.

LukasFrankenQ:
Hi @euronion, Hi @FabianHofmann,
Many thanks for the feedback and the suggestions, it is now all implemented. I also unified the docstrings, and think they are reasonable at explaining how to use the method. 
Please let me know if there are any further requests!

atlite/resource.py
```diff
@@ -34,31 +34,32 @@
...
-    turbine : str
-        Name of the local turbine file.
-        Alternatively a dict for selecting a turbine from the Open Energy
-        Database, in this case the key 'source' should be contained. For all
-        other key arguments to retrieve the matching turbine, see
-        atlite.resource.download_windturbineconfig() for details.
+    turbine : str or pathlib.Path
+        if str:
+            either from Open Energy Database (key "source" should then be contained)
+            or the name of a preshipped turbine from atlite.resources.windturbines
```
euronion, line 42:
```suggestion
        if str:
            The name of a preshipped turbine from atlite.resources.windturbines .
            Alternatively if a str starting with 'oedb:<name>' is passed the Open Energy
            Database is searched for a turbine with the matching '<name>' and if found
            that turbine configuration is used. See `atlite.resource.get_oedb_windturbineconfig(...)`
            for details.
```

atlite/resource.py
```diff
@@ -34,31 +34,32 @@
...
+        if Path: a local file is expected
```
euronion, line 43:
```suggestion
        if a `pathlib.Path` is provided the configuration is read from this local path instead.
```

atlite/resource.py
```diff
@@ -34,31 +34,32 @@
...
+    config : dict
+        Config with details on the turbine
     """
 
+    assert isinstance(turbine, (str, Path))
+
     if isinstance(turbine, str) and turbine.startswith("oedb:"):
         return get_oedb_windturbineconfig(turbine[len("oedb:") :])
 
     if isinstance(turbine, str):
```
euronion, line 56:
```suggestion
    elif isinstance(turbine, str):
```

atlite/resource.py
```diff
@@ -34,31 +34,32 @@
...
-        if not turbine.endswith(".yaml"):
-            turbine += ".yaml"
+        turbine_path = windturbines[turbine.replace(".yaml", "")]
 
-        turbine = WINDTURBINE_DIRECTORY / turbine
+    if isinstance(turbine, Path):
```
euronion, line 59:
```suggestion
    elif isinstance(turbine, Path):
```

atlite/resource.py
```diff
@@ -89,27 +106,29 @@ def get_cspinstallationconfig(installation):
...
 
+    assert isinstance(installation, (str, Path))
+
     if isinstance(installation, str):
-        if not installation.endswith(".yaml"):
-            installation += ".yaml"
+        installation_path = cspinstallations[installation]
 
-    installation = CSPINSTALLATION_DIRECTORY / installation
+    if isinstance(installation, Path):
```
euronion, line 125:
```suggestion
    elif isinstance(installation, Path):
```

atlite/resource.py
```diff
@@ -89,27 +106,29 @@ def get_cspinstallationconfig(installation):
...
     config : dict
         Config with details on the CSP installation.
     """
 
+    assert isinstance(installation, (str, Path))
+
     if isinstance(installation, str):
-        if not installation.endswith(".yaml"):
-            installation += ".yaml"
+        installation_path = cspinstallations[installation]
```
euronion, line 123:
```suggestion
        installation_path = cspinstallations[installation.replace(".yaml", "")]
```

atlite/resource.py
```diff
@@ -70,15 +71,31 @@ def get_windturbineconfig(turbine):
...
+        Config with details on the solarpanel
+    """
+
+    assert isinstance(panel, (str, Path))
 
     if isinstance(panel, str):
-        if not panel.endswith(".yaml"):
-            panel += ".yaml"
+        panel = panel.replace(".yaml", "")
+        panel_path = solarpanels[panel]
```
euronion, line 93:
```suggestion
        panel_path = solarpanels[panel.replace(".yaml", "")]
```

euronion, line 93:
```suggestion
        panel_path = solarpanels[panel.replace(".yaml", "")]
```

atlite/resource.py
```diff
@@ -70,15 +71,31 @@ def get_windturbineconfig(turbine):
...
 def get_solarpanelconfig(panel):
-    """Load the 'panel'.yaml file from local disk and provide a solar panel dict."""
+    """Load the 'panel'.yaml file from local disk and provide a solar panel dict.
+
+    Parameters
+    ----------
+    panel : str or pathlib.Path
+        if str: expects name to match one of the preshipped panels from
+            atlite.resources.solarpanel
+        if Path: a local file is expected
```
euronion, line 81: See windturbine

atlite/resource.py
```diff
@@ -89,27 +106,29 @@ def get_cspinstallationconfig(installation):
 
     Parameters
     ----------
-    installation : str
-        Name of CSP installation kind. Must correspond to name of one of the files
-        in resources/cspinstallation.
+    installation : str or pathlib.Path
+        if str: expects name to match one of the preshipped installations from
+            atlite.resources.cspinstallation
+        if Path: a local file is expected
```
euronion, line 112: See windturbine

euronion (approved):

PR #261, opened 2022-11-16
Combine ERA5 and ERA5T data

### LINKED ISSUE #190: ERA5 CDS requests which return a mixture of ERA5 and ERA5T data
## Description
As reported [here](https://confluence.ecmwf.int/pages/viewpage.action?pageId=173385064), CDS will return an dataset with an extra `expver ` dimension when a call spans over an date range that consist of both ERA5 and preliminary ERA5T data. This is not handled in Atlite. 

`expver 1` : ERA5 data, that has been quality checked. 
`expver 5` : ERA5T preliminary data (the three last months). 

## Expected Behavior
The  `expver ` dimension should be removed. 

A bonus would be to report the date when the ERA5T (`expver 5`)  data starts as this data might get updated/corrected later.

## Fix
An easy fix is that has worked in all my use cases:
`cutout.data = cutout.data.reduce(np.nansum, 'expver')`

Guess it could be done somewhere earlier in the preparation of the data, i.e. before any derived variables are calculated. Having the extra `expver` dimension also doubles the memory footprint, so it can be good to be thoughtful about this also when/if splitting cds calls in the time dimension

zoltanmaric:
Closes #190.

## Description
Requesting cutout data spanning recent (ERA5T) and data older than ~3 months (ERA5) results in an additional dimension in `cutout.data`, called `expver`, which `atlite` currently cannot handle gracefully.

This change collapses the two dimensions into a single dimension.

## Motivation and Context
See details in #190.

## How Has This Been Tested?
The first commit of this pull request introduces a failing test that demonstrates the bug. The second commit fixes the bug and makes the failing test pass.

~**IMPORTANT**: the test introduced in the first commit **should not be merged** into `master`. The test data is time-dependent. Data which is ERA5T today will become ERA5 in the future. I only added the test for illustration. See the discussion in this pull request for alternative test proposals.~ (This has been resolved)

## Type of change

## Checklist

### DIFF
--- a/test/test_preparation_and_conversion.py
+++ b/test/test_preparation_and_conversion.py
@@ -16,6 +16,8 @@ import sys
 import pytest
 import urllib3
 import geopandas as gpd
+from datetime import date
+from dateutil.relativedelta import relativedelta
 
 urllib3.disable_warnings()
 
@@ -42,7 +44,7 @@ def all_notnull_test(cutout):
 
 def prepared_features_test(cutout):
     """
-    The prepared features series should contain all variables in cuttout.data
+    The prepared features series should contain all variables in cutout.data
     """
     assert set(cutout.prepared_features) == set(cutout.data)
 
@@ -429,7 +431,7 @@ class TestERA5:
     @staticmethod
     def test_data_module_arguments_era5(cutout_era5):
         """
-        All data variables should have an attribute to which module thay belong
+        All data variables should have an attribute to which module they belong
         """
         for v in cutout_era5.data:
             assert cutout_era5.data.attrs["module"] == "era5"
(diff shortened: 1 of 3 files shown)


zoltanmaric:
> IMPORTANT: the test introduced in the first commit should not be merged into master. The test data is time-dependent. Data which is ERA5T today will become ERA5 in the future. I only added the test for illustration. See the discussion in this pull request for alternative test proposals.

@FabianHofmann This fix is addressing a "moving target" bug. The test case in the first commit spans September and October 2022, where the CDSAPI currently returns `ERA5` data for September, and `ERA5T` data for October, but this will shift in December ([source](https://confluence.ecmwf.int/pages/viewpage.action?pageId=173385064)).

In order to make this a stable test - we could keep the downloaded response data from CDSAPI in a file, and have the `retrieve_data` function load the request from file if the file is available. This would require another code change though. I made a dirty version of this solution in https://github.com/PyPSA/atlite/compare/master...zoltanmaric:atlite:reuse-downloaded-era5-files

I think this loading from file would be a useful feature anyway to prevent having to redownload large requests from the CDSAPI when debugging the building of cutouts.

test/test_preparation_and_conversion.py
```diff
@@ -306,6 +306,21 @@ def cutout_era5(tmp_path_factory):
...
+def cutout_mixed_era5_and_era5t(tmp_path_factory):
+    # Try just getting Berlin
+    # Try just requesting era5 and era5t
+    tmp_path = tmp_path_factory.mktemp("era5_era5t")
+    cutout = Cutout(
+        path=tmp_path / "era5_era5t",
+        module="era5",
+        bounds=BOUNDS,
+        time=slice("2022-09-01", "2022-10-31"),
+    )
```
FabianHofmann, line 319: Could we do something like this to address the moving target? 
```suggestion
    import datetime
    from datetime import date, timedelta
    today = date.today()
    first = today.replace(day=1)
    last_day_last_month = first - timedelta(days=1)
    first_day_second_last_month = (last_day_last_month - timedelta(days=32)).replace(day=1)

    tmp_path = tmp_path_factory.mktemp("era5_era5t")
    cutout = Cutout(
        path=tmp_path / "era5_era5t",
        module="era5",
        bounds=BOUNDS,
        time=slice(str(first_day_second_last_month), str(last_day_last_month)),
    )
```

FabianHofmann:
Thanks @zoltanmaric, always a pleasure to see contributions from you (sorry I am not so responsive at the `atlite` repo lately).  This looks very sensible.

test/test_preparation_and_conversion.py
zoltanmaric: We could, yeah. I was thinking it would lead to occasional continuous integration failures on the first day of each month, depending on when exactly the CDSAPI swaps `ERA5T` for `ERA5` data. But on second thought - the test would actually not fail in that case. Worst case it would either only contain `ERA5T` or only `ERA5` data - thus temporarily not testing the merging of the 2 datasets, but that's really not a big problem.

zoltanmaric: I updated the test accordingly - but reduced the time span from the full 2 months to 2 days (last of the second-previous month and first of the previous month) to make the CDSAPI request simpler and faster. Please have a look.

FabianHofmann:
@zoltanmaric sorry again that I am so slow here. Looks awesome!

PR #281, opened 2023-03-17
add option for zlib compression in .nc file cutout exports

fneum:


euronion:
Rejected

euronion:
I've directly commited a simpler version into your branch:

* without the additional function (not needed) and the ability to specify compression levels for an individual feature rather than re-applying the compression to all the data (potentially requiring recompression of existing features)
* Also: enabled by default (because it doesn't hurt)
* Removed additional function
* `to_file(..)` continues to work this way and not further code changes necessary :)

fneum:
I approve.

PR #290, opened 2023-04-28
era5: back extension to 1940

fneum:
https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-single-levels?tab=overview


fneum:
Confirmed, it works.

PR #255, opened 2022-09-14
solar pv tracking option added

enrilizaga:
## Change proposed in this Pull Request

Added 1-axis vertical and 2-axis tracking option for solar pv (+ small documentation addition for get_windturbineconfig)

## Description
Updated orientation.py and convert.py:
- Default option is no tracking (tracking = 0)
- For vertical tracking, surface azimuth = sun_azimuth angles 
- For 2-axis tracking, both the azimuth and the slope follow the sun, thus cos(incidence angle)=1

## Motivation and Context
This change adds the solar pv tracking capability to the code. Solar pv tracking is widely used as it increases the solar pv efficiency significantly. This change improves the fidelity of the code to represent those efficiencies

## How Has This Been Tested?
The updated code was tested locally (see snapshots)
- The following commands provided the same results (tracking = 0 default)
cap_factors_pv_notrack = cutout.pv(panel='CSi', orientation={'slope': 30., 'azimuth': 180.}, capacity_factor=True)
cap_factors_pv_0axis = cutout.pv(panel='CSi', orientation={'slope': 30., 'azimuth': 180.}, tracking=0, capacity_factor=True)
0tracking_comparison
0tracking_comparison_2

- Tracking = 1 and =2 were tested as well:
cap_factors_pv_1axis = cutout.pv(panel='CSi', orientation={'slope': 30., 'azimuth': 180.}, tracking=1, capacity_factor=True)
cap_factors_pv_2axis = cutout.pv(panel='CSi', orientation={'slope': 30., 'azimuth': 180.}, tracking=2, capacity_factor=True)
1tracking
2tracking

- Checked that csp was not affected (tracking change only for solar pv)
cf = cutout.csp(installation="SAM_solar_tower", capacity_factor=True)
cf = cutout.csp(installation="SAM_parabolic_trough", capacity_factor=True)
csp

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] I locally ran `pytest` inside the repository and no unexpected problems came up.
- [ ] I have adjusted the docstrings in the code appropriately.
- [ ] I have documented the effects of my code changes in the documentation `doc/`.
- [ ] I have added newly introduced dependencies to `environment.yaml` file.
- [ ] I have added a note to release notes `doc/release_notes.rst`.
[...]

PR #293, opened 2023-05-02
[Reviewed] PV horizontal tracking

fneum:
Same as #255 but with one additional commit.

Cleaned-up example notebook.


FabianHofmann (approved):
Would be nice to have a bit more sensible plot in the example notebook. Perhaps something like 
1. one summer week
2. yearly total

fneum:
Great @FabianHofmann! Let's tackle it later. The graphic and average capacity factors for May already illustrate the idea. 

Thanks @martavp @Parisra! That is a super nice feature!

PR #296, opened 2023-05-02
update deploy.yml

FabianHofmann:
copied from pypsa repo


fneum (approved):

FabianHofmann:
I am trying first with the current deploy.yml, and if every thing works well, we don't need this

PR #328, opened 2023-10-24
Add helper function ``area`` and ``layout_from_area_density`` & allow auto chunks

FabianHofmann:
- add functions `area` & `layout_from_area_density`
 - allow for `auto` chunks when loading cutout

## Motivation and Context

The ``area`` function is helpful for custom workflows, the same with the new layout function.

The auto chunking **can** help to accelerate the computation of the conversion functions. 

## Type of change

## Checklist

### DIFF
--- a/atlite/cutout.py
+++ b/atlite/cutout.py
@@ -575,7 +575,7 @@ class Cutout:
         ----------
         crs : int, optional
             The coordinate reference system (CRS) to use for the calculation.
-            The default value is 3035, which is a suitable projection for Europe and returns the area in square meters.
+            Defaults to the crs of the cutout.
 
         Returns
         -------
@@ -607,7 +607,7 @@ class Cutout:
             Capacity density in capacity/projection unit squared.
         crs : int, optional
             CRS to calculate the total area of grid cells.
-            Defaults to the crs of the cutout.
+            Passed to `Cutout.area()`.
 
         Returns
         -------
--- a/atlite/cutout.py
+++ b/atlite/cutout.py
@@ -150,7 +150,10 @@ class Cutout:
 
         path = Path(path).with_suffix(".nc")
         chunks = cutoutparams.pop("chunks", {"time": 100})
-        storable_chunks = {f"chunksize_{k}": v for k, v in (chunks or {}).items()}
+        if isinstance(chunks, dict):
+            storable_chunks = {f"chunksize_{k}": v for k, v in (chunks or {}).items()}
+        else:
+            storable_chunks = {}
 
         # Backward compatibility for xs, ys, months and years
         if {"xs", "ys"}.intersection(cutoutparams):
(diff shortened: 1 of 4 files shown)


atlite/cutout.py
```diff
@@ -564,6 +567,27 @@ def intersectionmatrix(self, shapes, shapes_crs=4326):
         """
         return compute_intersectionmatrix(self.grid, shapes, self.crs, shapes_crs)
 
+    def area(self, crs=3035):
```
euronion, line 570: I don't like the default `crs=3035` - it's only accurate for Europe.

By default I think it should use the CRS of the cutout or throw an error (asking the user to specify a CRS).

atlite/cutout.py
```diff
@@ -618,6 +642,27 @@ def layout_from_capacity_list(self, data, col="Capacity"):
...
+        """
+        Get a capacity layout from a uniform capacity density.
+
+        Parameters
+        ----------
+        capacity_density : float
+            Capacity density in capacity/km^2
+        crs : int, optional
+            CRS to calculate the total area of grid cells in m^2.
+            The default is 3035 which is suitable for the European area.
```
euronion, line 655: See comment in the `area` function on (default) CRS above .

atlite/cutout.py
```diff
@@ -618,6 +642,27 @@ def layout_from_capacity_list(self, data, col="Capacity"):
         )
         return data.to_xarray().reindex_like(self.data).fillna(0)
 
+    def layout_from_area_density(self, capacity_density, crs=3035):
+        """
+        Get a capacity layout from a uniform capacity density.
```
euronion, line 647: I find the formulation "from density" a bit confusing. What do you think about `uniform_density_layout(...)` instead, to be consistent with the `uniform_layout(...)`?

atlite/cutout.py
```diff
@@ -150,7 +150,10 @@ def __init__(self, path, **cutoutparams):
 
         path = Path(path).with_suffix(".nc")
         chunks = cutoutparams.pop("chunks", {"time": 100})
-        storable_chunks = {f"chunksize_{k}": v for k, v in (chunks or {}).items()}
+        if isinstance(chunks, dict):
+            storable_chunks = {f"chunksize_{k}": v for k, v in (chunks or {}).items()}
```
euronion, line 154: Is auto-chunking stable enough nowadays to become the default?

euronion (changes requested):
Minor comments

atlite/cutout.py
FabianHofmann: I don't think we should set it as default yet. There might still be some problems with `atlite` functionalities like reprojecting, which do not fully work with chunked spatial dimensions.

atlite/cutout.py
FabianHofmann: This actually a good point, thanks for raising. I adjusted the code accordingly.

atlite/cutout.py
coroa: `(chunks or {})` is equivalent to `chunks` (since `chunks` is a dict)

PR #346, opened 2024-04-12
Fix capacity factor timeseries in convert.py

TimFuermann:
The method convert and aggregate was adopted, so the capacity factor timeseries is returned and all functionality, such as return capacity and per_unit are kept.

Closes # (if applicable).

## Type of change

## Checklist

- [ ] I tested my contribution locally and it seems to work fine.

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -107,7 +107,8 @@ def convert_and_aggregate(
     resource : xr.DataArray
         Time-series of renewable generation aggregated to buses, if
         `matrix` or equivalents are provided else the total sum of
-        generated energy.
+        generated energy. If capacity_factor flag is set the capacity
+        factor for each grid cell is returned instead.
     units : xr.DataArray (optional)
         The installed units per bus in MW corresponding to `layout`
         (only if `return_capacity` is True).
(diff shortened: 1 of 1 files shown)


FabianHofmann:
thanks for the PR @TimFuermann, I will try to have a look at it soon

FabianHofmann:
Thanks again @TimFuermann, this makes really sense. The function itself is a mess, and it builds one critical core of all our analyses, which is why I postponed the review so long. But it looks good. I would like to have a second pair of eyes looking at it. @lkstrp could you also have a look if the main logics are preserved? It is mainly tracking the action steps within the function being subject to the set of arguments... there is no time pressure on this

lkstrp (approved):
One more sorry that this took so long @TimFuermann ! I full on lost that one.
But `atlite` will get more love from now on again.

I'll merge the fixed pre-commit and tests first, but this looks good.

Thanks you!

atlite/convert.py
```diff
@@ -119,22 +120,24 @@ def convert_and_aggregate(
...
+        if capacity_factor_timeseries:
+            results = da.rename("capacity factor")
         else:
-            res = da.sum("time", keep_attrs=True)
-            return maybe_progressbar(res, show_progress, **dask_kwargs)
+            results = da.mean("time").rename("capacity factor")
+        results.attrs["units"] = "p.u."
+
+    else:
+        results = da.sum("time", keep_attrs=True)
```
lkstrp, line 140: Removing the return means that we no longer end the function here, which leads to undefined issues later. I know, this one is open for quite a while now, but could you have another look?

Otherwise this function is quite a mess and could get a general refactoring with your proposed changes included

lkstrp (changes requested):

atlite/convert.py
TimFuermann: @lkstrp: I need to check it locally, but could you explain which undefined issues will occur later? I know the fix is not ideal, and was actually just done that kind of messy to not restructure the whole function.

lkstrp: Further down, `matrix` is not defined. If you merge the current master, this will hopefully show up in the tests. They may not run through because of a missing cdsapi token, as this branch was derived earlier. But you can run them locally in any case using `pytest .`.

``` python
if index is None:
    index = pd.RangeIndex(matrix.shape[0])
```

But as you said, restructuring the whole function is better. If you have already looked at it, feel free. Otherwise I will deal with it at some point.

TimFuermann: @lkstrp I restructured the function and tested it. I commit it soon. Nevertheless, this would need to change some things in the origional test files, too and in the usage. I explain it with the commit, Best, Tim

lkstrp: Sounds great! Thank you @TimFuermann

TimFuermann: @lkstrp Just uploaded a new version of how the convert and aggregate could be used more consistently. Nevertheless this includes heavy changes in the input flags, but allows for a more consistent way of using them. A default capacity layout is used if no capacities are provided. This is a constant 1MW per gridcell (which should be the same as looking for capacity factors). With this new implementation, capacity factors for regions, per gridcell, aswell as aggregated can be calculated.

lkstrp: @TimFuermann Still haven't found time do dive deeper in this which I need to do. But I will get back to you!

PR #352, opened 2024-07-01
added variable selection to make it compatible with sarah3

matzech:
I noticed that the code is not compatible with the novel SARAH-3 dataset as the input data has changed slightly.When reading with the existing code, it results into multiple variables which raises errors in the downstream conversion. Therefore, this fork propose to directly select the respective SIS and SID values making it compatible with the old and the new SARAH format. It uses double brackets to keep the same data type (xr.DataSet) as so far.

## Description
Only changed the lines that read the SARAH netcdf files by directly selecting the data variables.

## Motivation and Context
I had an issue opening SARAH -3 data with the existing code.

## How Has This Been Tested?
I tested it with the old SARAH -2 data and the novel SARAH-3. Both works as expected. I was not sure if I should add unit tests because there are, as far as I have seen, no unit tests for the different datasets so far. 

## Type of change
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist
- [x ] I tested my contribution locally and it seems to work fine.


fneum:
LGTM, maybe add a short release note announcing compatibility with SARAH-3. Unit test not necessary since we have none for SARAH-2 so far. Does the selection with double brackets preserve lazy loading?

fneum (approved):

matzech:
Thanks! The lazy loading behavior from dask is as in the status quo as variable indexing by names should not trigger .load() or .compute()

euronion:
So simple, so elegant. Thank you!

PR #356, opened 2024-07-10
fix: disable_progressbar in compute_availabilitymatrix

Irieo:

The function `compute_availabilitymatrix()` includes an option to disable the progress bar (`disable_progressbar`). However, for non-parallel computation path (i.e., when `nprocesses` is `None`, the code uses `tqdm` directly, w/o passing the `disable_progressbar` check.  This PR fixes it. User is informed if progress bar is disabled.

Why bother: it seems that **progress bar makes computation much slower**. For a test case (UA, solar availability matrix, local computation, nprocesses is None) disabling progress bar speeds the process from **2000 seconds to 760 seconds.** 
Similar behavior was observed for other atlite processes, such as `build_renewable_profiles` (rumor evidence).

Closes #[pypsa-eur/1090](https://github.com/PyPSA/pypsa-eur/issues/1090)

## Checklist

### DIFF
--- a/atlite/gis.py
+++ b/atlite/gis.py
@@ -727,10 +727,27 @@ def compute_availabilitymatrix(
         total=len(shapes),
         desc="Compute availability matrix",
     )
+
     if nprocesses is None:
+        logger.info(
+            "You can disable the progress bar by setting 'disable_progressbar=True'."
+        )
+
+        if not disable_progressbar:
+            tqdm_kwargs = dict(
+                ascii=False,
+                unit=" gridcells",
+                total=len(shapes),
+                desc="Compute availability matrix",
+            )
+            iterator = tqdm(shapes.index, **tqdm_kwargs)
+        else:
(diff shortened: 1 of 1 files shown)


atlite/gis.py
```diff
@@ -727,10 +727,27 @@ def compute_availabilitymatrix(
...
+                ascii=False,
+                unit=" gridcells",
+                total=len(shapes),
+                desc="Compute availability matrix",
+            )
+            iterator = tqdm(shapes.index, **tqdm_kwargs)
+        else:
+            logger.info("Progress bar is disabled.")
+            iterator = shapes.index
+
```
FabianHofmann, line 747:
```suggestion
        if not disable_progressbar:
            iterator = tqdm(shapes.index, **tqdm_kwargs)
        else:
            iterator = shapes.index
```
Great @Irieo , I think we can make it more minimalistic right?

fneum: Yes!

PR #364, opened 2024-08-02
era5: update to new CDS infrastructure

### LINKED ISSUE #361: Migrate to new CDS infrastructure
See announcement https://forum.ecmwf.int/t/the-new-climate-data-store-beta-cds-beta-is-now-live/3315

Migration to complete by September 2024.

fneum:
closes #361 

This PR is backwards compatible; old CDS infrastructure can still be used.

Other things needed to update:

- follow instructions https://forum.ecmwf.int/t/the-new-climate-data-store-beta-cds-beta-is-now-live/3315
- retrieve new `~/.cdsapirc`
- update to `cdsapi>=0.7.0`

Main changes:
- The time variable was renamed in CDS from "time" to "valid_time", which is renamed in `_rename_and_clean_coords()`
- A useless "number" coordinate was added, which is removed in `_rename_and_clean_coords()`.
- There was some weirdness in the "expver" coordinate when ERA5 and ERA5T data were mixed (coordinate but not a dimension). Also, it seems that the "expver" coordinate is now always created and is a 4-digit string rather than an integer. If somebody has a leaner suggestion here, I am all ears. It was quite tricky as non-`xarray` expert. 

If we want to use new CDS infrastructure in CI, the secrets need to be updated. Perhaps we shouldn't do this immediately.

Furthermore, I am not sure how to remove the plentiful INFO logs.

### DIFF
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -86,6 +86,8 @@ def _rename_and_clean_coords(ds, add_lon_lat=True):
     longitude columns as 'lat' and 'lon'.
     """
     ds = ds.rename({"longitude": "x", "latitude": "y"})
+    if "valid_time" in ds.sizes:
+        ds = ds.rename({"valid_time": "time"}).unify_chunks()
     # round coords since cds coords are float32 which would lead to mismatches
     ds = ds.assign_coords(
         x=np.round(ds.x.astype(float), 5), y=np.round(ds.y.astype(float), 5)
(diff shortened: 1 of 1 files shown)


FabianHofmann (approved):
Don't understand it fully, but looks good

euronion:
* Should restrict to new CDSAPI package version
* Should add migration note in the release notes
* Should make a new release

I think migrating ASAP makes sense, since the old one will be shutdown in less than a month.

Issue #363, opened 2024-08-01
`compute_shape_availability` triggers numpy failure: Python integer 255 out of bounds for int8


irm-codebase:
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/atlite/releases) of Atlite.

- [ ] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/atlite/tree/master) branch of Atlite.

### Issue Description

While trying to reproduce the land availability example, the following error is returned.

It seems like this is caused by a depreciated functionality in `numpy`: https://github.com/numpy/numpy/issues/26596

Either `numpy` or something else needs to be pinned in the environment, maybe?

```
File ~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:586, in ExclusionContainer.compute_shape_availability(self, geometry, dst_transform, dst_crs, dst_shape)
    [582](https://file+.vscode-resource.vscode-cdn.net/home/ivanruizmanuel/Documents/git/ec_modules/tmp/~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:582)     return shape_availability_reprojected(
    [583](https://file+.vscode-resource.vscode-cdn.net/home/ivanruizmanuel/Documents/git/ec_modules/tmp/~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:583)         geometry, self, dst_transform, dst_crs, dst_shape
    [584](https://file+.vscode-resource.vscode-cdn.net/home/ivanruizmanuel/Documents/git/ec_modules/tmp/~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:584)     )
    [585](https://file+.vscode-resource.vscode-cdn.net/home/ivanruizmanuel/Documents/git/ec_modules/tmp/~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:585) else:
--> [586](https://file+.vscode-resource.vscode-cdn.net/home/ivanruizmanuel/Documents/git/ec_modules/tmp/~/miniforge3/envs/ec-atlite-pv-series/lib/python3.12/site-packages/atlite/gis.py:586)     return shape_availability(geometry, self)
...
[...]

PR #365, opened 2024-08-02
temporarily limit numpy version to <2

fneum:
xref #363 

and also some tests were super slow

to be investigated


FabianHofmann:
ah unfortunate... thanks @fneum

FabianHofmann (approved):

PR #367, opened 2024-08-05
use dask.array functions instead of numpy

fneum:
I had hoped for some performance gains, but this is not so clear.


euronion:
NB:
Most of the `dask` functions you have changed are just calls to the same `numpy` functions using `ufunc`, see the "direct" entries here: https://docs.dask.org/en/stable/array-numpy-compatibility.html#compatibility-with-numpy-functions .

Changing does give a bit of consistency, I'm not opposed.

PR #375, opened 2024-08-17
CI: use CDS BETA

fneum:
Secret `CDSAPI_TOKEN_BETA`has been added to the repository.


fneum:
old CDS API is going to be shut down in two weeks, on September 26

> The [legacy CDS](https://cds.climate.copernicus.eu/#!/home) and [legacy ADS](https://ads.atmosphere.copernicus.eu/cdsapp#!/home) will be decommissioned on September 26, 2024 and will no longer be accessible from this date onwards.

https://forum.ecmwf.int/

PR #236, opened 2022-05-16
Enable download of large (spatial extent) cutouts from ERA5 via cdsapi.

euronion:
Closes #221 .

## Change proposed in this Pull Request

Split download of ERA5 into monthly downloads (currently: annual downloads) to prevent too-large downloads from ERA5 CDSAPI.

## TODO

- [x] Add month indicator to progress prompts.

## Motivation and Context
See #221 .

## How Has This Been Tested?
Locally by downloading a large cutout.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [n/a] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] I tested my contribution locally and it seems to work fine.
- [x] I locally ran `pytest` inside the repository and no unexpected problems came up.
- [x] I have adjusted the docstrings in the code appropriately.
- [x] I have documented the effects of my code changes in the documentation `doc/`.
- [n/a] I have added newly introduced dependencies to `environment.yaml` file.
- [x] I have added a note to release notes `doc/release_notes.rst`.
- [x] I have used `pre-commit run --all` to lint/format/check my contribution

PR #401, opened 2024-11-01
fix: Skip previous encoding workaround for fixed xarray versions

coroa:
Follow up to a workaround introduced by @euronion in #236 for an encoding issue in xarray (https://github.com/pydata/xarray/issues/7691), which was fixed by (https://github.com/pydata/xarray/pull/8713) and eventually included xarray v2024.03.0 (https://github.com/pydata/xarray/releases/tag/v2024.03.0).

## Checklist

### DIFF
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -12,6 +12,7 @@ import logging
 import os
 import warnings
 import weakref
+from importlib.metadata import version
 from tempfile import mkstemp
 
 import cdsapi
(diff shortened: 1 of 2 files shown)


lkstrp:
We could also just pin 'xarray' now

coroa:
> We could also just pin xarray now

I guess with all the changes xarray is undergoing, having a half-year old version is lower bound is quite acceptable, true.

euronion:
Thanks for the cleanup. I'm also in favour of restricting the `xarray` version. Btw. did you test whether the fix in `xarray` actually worked? I never got to testing it...

coroa:
> Thanks for the cleanup. I'm also in favour of restricting the `xarray` version. Btw. did you test whether the fix in `xarray` actually worked? I never got to testing it...

The dataarrays we receive from CDS has in xarray an encoding of `float32` now. As far as i understand that means the fix is working.

coroa:
Ok, i replaced the implementation by a new version pin.

lkstrp (approved):

PR #402, opened 2024-11-01
feat(wind): Add power law interpolation method for wind conversion

### LINKED ISSUE #231: Use power low instead of log law for extrapolating wind speed
Use power low instead of log law for extrapolating wind speed to hub height.

## Description
The `extrapolate_wind_speed` function uses the power law and the grid averages forecast surface roughness (fsr):
 `wnd_spd = ds[from_name] * (   np.log(to_height / ds["roughness"]) / np.log(from_height / ds["roughness"])  )`
Using the ERA5 fsr with the log law results in too large correction, especially in the Nordics where the fsr is often high.  
At height above the blending height it is better to use the power law. That looks like follow:
`wnd_spd = ds["wnd100m"] * (to_height / 100)**a`. 
The wind sheer exponent `a` can be derived from the wind speed at 10m and 100m: 
`a = np.log(ds["wnd10m"] / ds["wnd100m"]) /  np.log(10 / 100)`. 
Tested it and it give same results as doing actual vertical interpolations with MetPy using the wind components from the pressure level data (how the ERA5 wind at 100 meters is orginally calculated as well).  

For turbines hub heights lower than the blending height it could be useful to use the log law to account for site specific roughness length. Below the equations from the IFS CY41R2 docs used to derive the 10m wind speed in ERA5 (always calculated assuming fsr of 0.03 to better represent conditions at weather stations)

image
The IFS docs mention two blending heights, 40 and 75 meters (I guess it actually might be dynamic and site specific). But a procedure I use my self is to first use the above-mentioned power law to calculate a wind speed at a blending height of 70 meters, then the log law using a site specific fsr (for now, just some blunt estimates based on the ERA5 grid average fsr)

coroa:
Closes #231 .

Refer to issue for context. Implements the power law interpolation method as a new argument
to `cutout.wind`. Needs to be chosen manually for now.

A suggested in the issue we retrieve 100m and 10m wind speeds and then calculate a wind shear exponent from those during cutout preparation.

Looks like the power law is preferred at least for turbine heights above wind blending (?) height (70m?). So this should most likely become the new default if we have 100m wind speeds available (as we have for ERA5). 

TODOs:
- [ ] Compare results, maybe following methodology in https://www.mdpi.com/1996-1073/14/14/4169
- [ ] Implement a FutureWarning that the default changes to the power law.
- [ ] Add those interpolation options to the docs.

## Checklist

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -478,19 +478,21 @@ def solar_thermal(
 
 
 # wind
-def convert_wind(ds, turbine):
+def convert_wind(ds, turbine, interpolation_method):
     """
     Convert wind speeds for turbine to wind energy generation.
     """
     V, POW, hub_height, P = itemgetter("V", "POW", "hub_height", "P")(turbine)
 
-    wnd_hub = windm.extrapolate_wind_speed(ds, to_height=hub_height)
+    wnd_hub = windm.extrapolate_wind_speed(
+        ds, to_height=hub_height, method=interpolation_method
+    )
 
-    def _interpolate(da):
+    def apply_power_curve(da):
         return np.interp(da, V, POW / P)
 
     da = xr.apply_ufunc(
-        _interpolate,
+        apply_power_curve,
         wnd_hub,
         input_core_dims=[[]],
         output_core_dims=[[]],
@@ -503,7 +505,14 @@ def convert_wind(ds, turbine):
     return da
 
 
-def wind(cutout, turbine, smooth=False, add_cutout_windspeed=False, **params):
+def wind(
+    cutout,
+    turbine,
+    smooth=False,
+    add_cutout_windspeed=False,
+    interpolation_method="logarithmic",
+    **params,
+):
     """
     Generate wind generation time-series.
 
@@ -529,6 +538,9 @@ def wind(cutout, turbine, smooth=False, add_cutout_windspeed=False, **params):
         output at the highest wind speed in the power curve. If False, a warning will be
         raised if the power curve does not have a cut-out wind speed. The default is
         False.
+    interpolation_method : {"logarithmic", "power"}
+        Law to interpolate wind speed to turbine hub height. Refer to docstring of
+        atlite.wind.extrapolate_wind_speed .
 
     Note
     ----
@@ -550,7 +562,10 @@ def wind(cutout, turbine, smooth=False, add_cutout_windspeed=False, **params):
         turbine = windturbine_smooth(turbine, params=smooth)
 
     return cutout.convert_and_aggregate(
-        convert_func=convert_wind, turbine=turbine, **params
+        convert_func=convert_wind,
+        turbine=turbine,
+        interpolation_method=interpolation_method,
+        **params,
     )
 
 
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -539,7 +539,7 @@ def wind(
         raised if the power curve does not have a cut-out wind speed. The default is
         False.
     interpolation_method : {"logarithmic", "power"}
-        Law to interpolate wind speed to turbine hub height. Refer to 
+        Law to interpolate wind speed to turbine hub height. Refer to
         :py:func:`atlite.wind.extrapolate_wind_speed`.
 
     Note
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -539,8 +539,8 @@ def wind(
         raised if the power curve does not have a cut-out wind speed. The default is
         False.
     interpolation_method : {"logarithmic", "power"}
-        Law to interpolate wind speed to turbine hub height. Refer to docstring of
-        atlite.wind.extrapolate_wind_speed .
+        Law to interpolate wind speed to turbine hub height. Refer to 
+        :py:func:`atlite.wind.extrapolate_wind_speed`.
 
     Note
     ----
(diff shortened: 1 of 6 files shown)


coroa:
Tests work locally, any pointers to what is going wrong in CI? @lkstrp

lkstrp:
This PR doesn't have permissions to get the CDS API secret because it comes from a fork, which causes these errors. We might want to change the trigger, or add a manual trigger that has permissions. But the Cutouts should also just be fully cached.

atlite/convert.py
```diff
@@ -529,6 +538,9 @@ def wind(cutout, turbine, smooth=False, add_cutout_windspeed=False, **params):
         output at the highest wind speed in the power curve. If False, a warning will be
         raised if the power curve does not have a cut-out wind speed. The default is
         False.
+    interpolation_method : {"logarithmic", "power"}
+        Law to interpolate wind speed to turbine hub height. Refer to docstring of
+        atlite.wind.extrapolate_wind_speed .
```
lkstrp, line 543:
```suggestion
        Law to interpolate wind speed to turbine hub height. Refer to 
        :py:func:`atlite.wind.extrapolate_wind_speed`.
```

atlite/wind.py
```diff
@@ -13,14 +13,18 @@
...
+):
     """
     Extrapolate the wind speed from a given height above ground to another.
 
     If ds already contains a key refering to wind speeds at the desired to_height,
     no conversion is done and the wind speeds are directly returned.
 
-    Extrapolation of the wind speed follows the logarithmic law as desribed in [1].
+    Extrapolation of the wind speed can either use the "logarithmic" law as
+    described in [1] or the "power" law as described in [2]. See also discussion
```
lkstrp, line 26:
```suggestion
    described in [1]_ or the "power" law as described in [2]_. See also discussion
```

atlite/wind.py
```diff
@@ -47,9 +53,9 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
     time series from a new global renewable energy atlas for energy system
     analysis'.
```
lkstrp, line 54:
```suggestion
    .. [1] Equation (2) in Andresen, G. et al (2015): 'Validation of Danish wind
       time series from a new global renewable energy atlas for energy system
       analysis'.
```

atlite/wind.py
```diff
@@ -47,9 +53,9 @@ def extrapolate_wind_speed(ds, to_height, from_height=None):
...
 
-    [2] https://en.wikipedia.org/w/index.php?title=Roughness_length&oldid=862127433,
-    Retrieved 2019-02-15.
-
+    [2] Gualtieri, G. (2021): 'Reliability of ERA5 Reanalysis Data for Wind
+    Resource Assessment: A Comparison against Tall Towers'
+    https://doi.org/10.3390/en14144169 .
```
lkstrp, line 58:
```suggestion
    .. [2] Gualtieri, G. (2021): 'Reliability of ERA5 Reanalysis Data for Wind
       Resource Assessment: A Comparison against Tall Towers'
       https://doi.org/10.3390/en14144169 .
```

lkstrp (approved):
In `pypsa` and `linopy` we now use static types. So always feel free to spam those in there if you touch anything. Long way to go for `atlite` though.

And great to have you back 🎉

atlite/datasets/era5.py
```diff
@@ -44,7 +44,7 @@ def nullcontext():
 
 features = {
     "height": ["height"],
-    "wind": ["wnd100m", "wnd_azimuth", "roughness"],
+    "wind": ["wnd100m", "wnd_shear_exp", "wnd_azimuth", "roughness"],
```
fneum, line 47: Any problems if the new extrapolation functions are used with an old cutout?

fneum (commented):
I like it!

coroa:
> In `pypsa` and `linopy` we now use static types. So always feel free to spam those in there if you touch anything. Long way to go for `atlite` though.

Added quite some typing to the wind conversion now. Needs `typing-extensions` and `from __future__ import annotations` though.

Hope that is agreed practice.

atlite/datasets/era5.py
coroa: You would raise a `KeyError` when `atlite.wind.extrapolate_wind_speed` tries to access `ds["wnd_shear_exp"]`.

We could raise a more informative error message instead, i guess.

lkstrp:
> Added quite some typing to the wind conversion now. Needs `typing-extensions` and `from __future__ import annotations` though.
> 
> Hope that is agreed practice.

Great, it is. The `__future__` import is only needed for Python `3.9`, and we can remove it once we deprecate `3.9` (and  support `3.13` instead).

atlite/datasets/era5.py
coroa: Done now

coroa:
> > Added quite some typing to the wind conversion now. Needs `typing-extensions` and `from __future__ import annotations` though.
> > Hope that is agreed practice.
> 
> Great, it is. The `__future__` import is only needed for Python `3.9`, and we can remove it once we deprecate `3.9` (and support `3.13` instead).

Yeah, well, i am confused about the state of the annotations :), since I started reading into https://docs.python.org/3.11/whatsnew/3.11.html#pep-563-may-not-be-the-future .

fneum (approved):
LGTM.

Side note: Would it make sense to at some point split the `wind_test` into multiple smaller tests?

PR #403, opened 2024-11-01
feat(wind): Bias correct wind speeds based on scaling to a known average

coroa:
The ERA5 long run average is determined during cutout preparation by averaging the monthly mean wind speeds from 1980 to the last year and made available as a dataset variable wnd100m_lra.

Example usage:

```python
cutout.wind(
    atlite.windturbines.Enercon_E101_3000kW,
    real_long_run_average_windspeed="gwa3_250_windspeed_100m.tif",
)
```

TODO:
- [ ] Add docs
- [ ] Add tests

Closes #373 .

## Checklist

- [ ] Code changes are sufficiently documented; i.e. new functions contain docstrings and further explanations may be given in `doc`.
- [ ] Unit tests for new features were added (if applicable).
- [ ] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).
- [ ] A note for the release notes `doc/release_notes.rst` of the upcoming release is included.
- [ ] I consent to the release of this PR's code under the MIT license.

PR #405, opened 2024-11-09
feat(wind): Bias correct wind speeds based on scaling to a known average (alternative)

### LINKED ISSUE #373: Allow wind speeds to be corrected by a location-specific linear factor (e.g. from GWA)
We should bias correct ERA5's wind speeds using DTU's global wind atlas (GWA). This will raise wind speeds in e.g. onshore Germany and the Alps.

Screenshot from 2024-08-14 18-06-29

See a [similar PR in PyPSA-Eur](https://github.com/PyPSA/pypsa-eur/pull/377) from @p-glaum.

Consensus was we should wait until ERA6 is published before taking this on, since ERA6 will have improved resolution.

# Implementation

Allow some sort of matrix mask to be passed to `convert_wind` that pre-multiplies the wind speeds in `convert.py`. Perhaps provide functionality/example for GWA, or leave that to PyPSA-X models.

coroa:
Alternative to #403 .

Adds a new function :py:func:`atlite.datasets.era5.retrieve_average_windspeed`
to retrieve average windspeeds. This dataset function is called by
:py:func:`atlite.wind.calculate_windspeed_bias_correction` to derive a bias
correction which can be passed to the default wind generation.

Example usage:

```python
windspeed_bias_correction = atlite.wind.calculate_windspeed_bias_correction(
    cutout,
    real_average="gwa3_250_windspeed_100m.tif",
    height=100,
)
cutout.wind(
    atlite.windturbines.Enercon_E101_3000kW,
    windspeed_bias_correction=windspeed_bias_correction,
    windspeed_height=100,
)
```

Closes #373 .

## Checklist

### DIFF
--- a/atlite/convert.py
+++ b/atlite/convert.py
@@ -478,19 +478,28 @@ def solar_thermal(
 
 
 # wind
-def convert_wind(ds, turbine):
+def convert_wind(
+    ds: xr.Dataset, turbine, windspeed_bias_correction=None, from_height=None
+):
     """
     Convert wind speeds for turbine to wind energy generation.
     """
     V, POW, hub_height, P = itemgetter("V", "POW", "hub_height", "P")(turbine)
 
-    wnd_hub = windm.extrapolate_wind_speed(ds, to_height=hub_height)
+    if windspeed_bias_correction is not None:
+        ds = ds.assign(
+            {f"wnd{from_height}m": ds[f"wnd{from_height}m"] * windspeed_bias_correction}
+        )
 
-    def _interpolate(da):
+    wnd_hub = windm.extrapolate_wind_speed(
+        ds, to_height=hub_height, from_height=from_height
+    )
+
+    def apply_power_curve(da):
         return np.interp(da, V, POW / P)
 
     da = xr.apply_ufunc(
-        _interpolate,
+        apply_power_curve,
         wnd_hub,
         input_core_dims=[[]],
         output_core_dims=[[]],
@@ -503,7 +512,15 @@ def convert_wind(ds, turbine):
     return da
 
 
-def wind(cutout, turbine, smooth=False, add_cutout_windspeed=False, **params):
+def wind(
+    cutout,
+    turbine: str | dict,
+    smooth: bool | dict = False,
+    add_cutout_windspeed: bool = False,
+    windspeed_bias_correction: xr.DataArray | None = None,
+    windspeed_height: int | None = None,
+    **params,
+):
     """
     Generate wind generation time-series.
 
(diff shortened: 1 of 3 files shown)


coroa:
@lkstrp Do you understand the missing schema errors in the macos for the data download?

lkstrp:
Ahh, we had that here in #433. Which was a big mystery to me as the PR doesn't even touch anything related

coroa:
Damn, the way i retrieve the long-run average of the era5 data does not work, since the monthly means are of the vector components. rather than the wind speed. which is necessarily lower.

there is probably no way around retrieving the full hourly 100m windspeeds for several years. i think we have to provide this as an external input.

coroa:
Ok, updated to use the hourly windspeeds and new usage modi which will allow us to distribute a global correction factor, but also embed it into the cutout.

atlite/convert.py
```diff
@@ -565,14 +565,19 @@ def convert_wind(
...
     turbine: TurbineConfig,
     interpolation_method: Literal["logarithmic", "power"],
+    windspeed_bias_correction: bool | xr.DataArray | None = None,
 ) -> xr.DataArray:
     """
     Convert wind speeds for turbine to wind energy generation.
     """
     V, POW, hub_height, P = itemgetter("V", "POW", "hub_height", "P")(turbine)
 
+    ds, from_height = windm.apply_windspeed_bias_correction(
```
lkstrp, line 575: The check if `apply_windspeed_bias_correction` actually does something should be done here and not within the method. If it's called it should do/ raise something or should not even be called.

atlite/data.py
```diff
@@ -31,14 +31,15 @@ def get_features(
     tmpdir=None,
     monthly_requests=False,
     concurrent_requests=False,
+    **parameter_updates,
```
lkstrp, line 34: We can use this for internal methods, but it's much more readable and gives a better docs/api reference if we avoid this general kwargs handling

atlite/datasets/era5.py
```diff
@@ -517,6 +524,120 @@ def retrieve_data(
...
+    ).mean("year")
+
+
+def get_data_windspeed_bias_correction(cutout, retrieval_params, creation_parameters):
+    """
+    Get windspeed bias correction
+    """
+    real_average_path = creation_parameters.get("windspeed_real_average_path")
+    if real_average_path is None:
+        logger.warning(
```
lkstrp, line 626: This warning is going to appear for every single cutout creation where the path is not specified, isn't it? Shouldn't we make it less verbose and more optional? The information should go in the docstring for the argument in `Cutout()` instead

atlite/datasets/era5.py
```diff
@@ -552,8 +673,10 @@ def get_data(
         If True, the monthly data requests are posted concurrently.
         Only has an effect if `monthly_requests` is True.
     **creation_parameters :
-        Additional keyword arguments. The only effective argument is 'sanitize'
-        (default True) which sets sanitization of the data on or off.
+        Additional keyword arguments.
```
lkstrp, line 676: As mentioned above

atlite/wind.py
```diff
@@ -115,14 +123,172 @@ def extrapolate_wind_speed(
...
+    uses either an explicitly provided scaling factor or a pre-calculated
+    scaling factor supplied in cutout.
+
+    Parameters
+    ----------
+    ds : xr.Dataset
+        Dataset containing wind speed time-series (at one or multiple heights)
+        with keys like 'wnd{height:d}m' and potentially a pre-calculated scaling
+        factor at key 'wnd_bias_correction'
+    windspeed_bias_correction : bool or DataArray, optional
```
lkstrp, line 240: I am also not a fan of these "could be anything" arguments. 
`False` is not needed, you just don't execute the function.
`None` uses a default from ds if it exists, otherwise does nothing is not clean either. This logic can be done before. You either use it or you don't by not calling this method. `None` can be used, but if you rely on a default it should raise an Error if the data for that does not exist. Same goes for `True`.

lkstrp (commented):
I only had a quick look and I have a couple of style things. But some are probably a general issue of the architecture of `atlite`, to which this PR fits well, and nothing that is introduced here. But I would like to keep it in mind for the history books/resolve at some point.

Can we add unit tests to all the new methods? I know the code cov diff is a pain, but every new method should have unit tests.

And how did you solve the MacOS runner problem?

atlite/data.py
coroa: Well, we don't know at cutout creation or in the prepare doc-string, which of these parameters exist, since they are datasets module specific. So, I don't think there is much i can do about this here.

atlite/datasets/era5.py
coroa: Cutout creation is not happening often and it is easy to miss this new feature. Further, it is new that the preparation of an available feature is skipped, when a data argument is missing, previously the implicit dataset generation rules meant that it should raise, but raising would be to invasive here.

Therefore, I decided for it to be verbose.

Same as above, it cannot go into the Cutout, since this is an era5 specific argument.

atlite/wind.py
coroa: The default has been chosen to not introduce any problems for existing uses.

atlite/datasets/era5.py
coroa: Inherent to the data preparation routines.

coroa:
Hi @lkstrp , thanks for the comments. I think most of your issues point at us having to overhaul the cutout generation and disentangle it from Cutout's `__init__` method.

atlite/datasets/era5.py
lkstrp: It is already in the [cutout init](https://github.com/PyPSA/atlite/pull/405/files/f1e154e7ec558b15f390b6d3d46a1c08d2b8713c#diff-3b908fa193d6a8b229f10313d0bb5f18c95441c9212f5e350143c24bb7fbd29cR137-R141), so the info should be added there as well.

I still disagree that this should be a default warning, but I see your points. If you want to go with it, mention in the warning that this is a new feature from `v0.5.0` and all docs would need to be updated

atlite/wind.py
lkstrp: But 2-4 can be handled before calling all those `apply_`... methods that may do nothing. 
It just makes it much harder to follow the code. Mypy will also be a pain to implement and less helpful with these many typed variables. I am not saying that this needs to be changed here, but in an ideal world we would have most of the variables in a single type. For sure internally

coroa: Ok, moved the argument treatment up into the cutout.wind method. Does this make your mypy life less painful?

atlite/datasets/era5.py
coroa: Ok, I am not married to the warning. Down-graded it to a debug message and copied the GWA reference also into the Cutout.__init__ docstring.

coroa:
@lkstrp I think I addressed your code style comments where adequate. Additional changes would require us to come up with a better cutout preparation routine.

I still need to add tests and update docs.

euronion:
@coroa I haven't looked at the new version yet - let me know when I should review.

Thanks for the great effort! :)

coroa:
> @coroa I haven't looked at the new version yet - let me know when I should review.
> 
> Thanks for the great effort! :)

I expect the code to be stable now for review.

Further required efforts are unit tests and docs, that i will start tackling starting next weeks.

jome1:
Is there an update? Is there anything I could help with? I am really looking forward to the bias-correction in atlite. It will be very helpful!

PR #414, opened 2024-12-03
update retrieve_data in era5.py

### LINKED ISSUE #413: Changed Dataformat from ERA5 Single Levels: .nc to .zip
### Version Checks (indicate both or one)

- [x] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/atlite/releases) of atlite.

- [X] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/atlite/tree/master) branch of atlite.

### Issue Description

Downloading era5 single-level-values from CDS has changed. In detail, the selection "data-format": "netcdf" returns a .zip file with multiple single .nc files by default, in case multiple fields are downloaded at once.

I suggest changing the retrieve_data function in the era5.py dataset, to per default download a .zip file and combine all .nc files within it directly.

The changes are linked in a few minutes.

### Reproducible Example

```python
import atlite

cutout = atlite.Cutout(
    path="western-europe-2011-01.nc",
    module="era5",
    x=slice(-13.6913, 1.7712),
    y=slice(49.9096, 60.8479),
    time="2011-01",
)

cutout.prepare()
```

### Expected Behavior

The expected behavior should be to return the cutout, instead the code fails, with an value error (see below). The value comes from the fact, that we try to load a .zip file as a .nc file and need to dezip it beforehand.

*** ValueError: did not find a match in any of xarray's currently installed IO backends ['netcdf4', 'scipy', 'rasterio']. Consider explicitly selecting one of the installed engines via the `engine parameter, or installing additional IO dependencies, see:
https://docs.xarray.dev/en/stable/getting-started-guide/installing.html
https://docs.xarray.dev/en/stable/user-guide/io.html

### Installed Versions

Local version of the master branch.
Added: IO Backend for Xarray -> ['h5netcdf']

TimFuermann:
- changed download_format to .zip
- added entcrypting of .zip file from download and merging of contained .nc files
- add url

Closes # (if applicable).

## Checklist

### DIFF
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -292,9 +292,9 @@ def retrieval_times(coords, static=False, monthly_requests=False):
     time = coords["time"].to_index()
     if static:
         return {
-            "year": [str(time[0].year)],
+            "year": [time[0].strftime("%Y")],
             "month": [time[0].strftime("%m")],
-            "day": [str(time[0].day).zfill(2)],
+            "day": [time[0].strftime("%d")],
             "time": time[0].strftime("%H:00"),
         }
 
@@ -306,19 +306,19 @@ def retrieval_times(coords, static=False, monthly_requests=False):
             for month in t.month.unique():
                 query = {
                     "year": str(year),
-                    "month": [str(month).zfill(2)],
+                    "month": [t[t.month == month][0].strftime("%m")],
                     "day": list(
-                        t[t.month == month].day.unique().astype(str).str.zfill(2)
+                        t[t.month == month].strftime("%d").unique()
                     ),
-                    "time": ["%02d:00" % h for h in t[t.month == month].hour.unique()],
+                    "time": list(t[t.month == month].strftime("%H:00").unique()),
                 }
                 times.append(query)
         else:
             query = {
                 "year": [str(year)],
-                "month": list(t.month.unique().astype(str).str.zfill(2)),
-                "day": list(t.day.unique().astype(str).str.zfill(2)),
-                "time": ["%02d:00" % h for h in t.hour.unique()],
+                "month": list(t.strftime("%m").unique()),
+                "day": list(t.strftime("%d").unique()),
+                "time": list(t.strftime("%H:00").unique()),
             }
             times.append(query)
     return times
--- a/atlite/datasets/era5.py
+++ b/atlite/datasets/era5.py
@@ -8,12 +8,12 @@ For further reference see
 https://confluence.ecmwf.int/display/CKB/ERA5%3A+data+documentation
 """
 
+import io
 import logging
 import os
-import io
-import zipfile
 import warnings
 import weakref
+import zipfile
 from tempfile import mkstemp
 
 import cdsapi
@@ -340,15 +340,17 @@ def retrieve_data(product, chunks=None, tmpdir=None, lock=None, **updates):
     If you want to track the state of your request go to
     https://cds-beta.climate.copernicus.eu/requests?tab=all
     """
-    
-    # Set url for data download, this allows to switch to different data 
+
+    # Set url for data download, this allows to switch to different data
     # sources more easily.
-    url = 'https://cds.climate.copernicus.eu/api'
-    
-    request = {"product_type": ["reanalysis"],
-               "data_format": "netcdf", 
-               "download_format": "zip"}
-    
+    url = "https://cds.climate.copernicus.eu/api"
+
+    request = {
+        "product_type": ["reanalysis"],
+        "data_format": "netcdf",
+        "download_format": "zip",
+    }
+
     request.update(updates)
 
     assert {"year", "month", "variable"}.issubset(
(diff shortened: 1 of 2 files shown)


TimFuermann:
This pull request should solve the issue mentioned in #414

atlite/datasets/era5.py
```diff
@@ -338,36 +340,76 @@ def retrieve_data(product, chunks=None, tmpdir=None, lock=None, **updates):
     If you want to track the state of your request go to
     https://cds-beta.climate.copernicus.eu/requests?tab=all
     """
-    request = {"product_type": "reanalysis", "format": "netcdf"}
+
+    # Set url for data download, this allows to switch to different data
+    # sources more easily.
+    url = "https://cds.climate.copernicus.eu/api"
```
lkstrp, line 346: How would you switch to other sources? Could you explain?
That should probably be a constant at module level, and maybe an optional argument

atlite/datasets/era5.py
```diff
@@ -338,36 +340,76 @@ def retrieve_data(product, chunks=None, tmpdir=None, lock=None, **updates):
...
+        url=url, info_callback=logger.debug, debug=logging.DEBUG >= logging.root.level
     )
     result = client.retrieve(product, request)
 
     if lock is None:
         lock = nullcontext()
 
     with lock:
-        fd, target = mkstemp(suffix=".nc", dir=tmpdir)
+        fd, target_zip = mkstemp(suffix=".zip", dir=tmpdir)
```
lkstrp, line 369: Do you have any experience on whether zipping always, even if `.nc' files could be downloaded, has any performance drawbacks?
I could imagine that it might even improve it, since the bottleneck is most likely download bandwidth, but I do not know.

pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "numexpr",
     "xarray>=2024.03.0",
     "netcdf4",
+    "h5netcdf",
```
lkstrp, line 39: Why is `h5netcdf` added? Can it not also replace `netcdf4`? Not sure if we wanna have both as dependency and not one of them optional.

lkstrp (commented):
Thank you @TimFuermann ! I just have a couple of questions

atlite/datasets/era5.py
TimFuermann: My experience was that downloading .zip files directly, improves performance. Nevertheless, I did not test it.

pyproject.toml
TimFuermann: h5netcdf is somehow required, as the datastructure changed. It seems that the nc files within the zip folder use a different compression. There is so much going on at the CDS,, not completly clear from my side.
One comment on the forum mentions the general problem here: https://forum.ecmwf.int/t/problems-parsing-nc-files-after-outage/8060/3

For me netcdf4 was not working, I had to isntall h5netcdf, I could not directly find it at the ECMW forum, but it could be worth trying it with an update of the netcdf4 converter.

atlite/datasets/era5.py
TimFuermann: Hi Lukas, you are right, I just added it here for later improvement. For example, I have a local version in which I added the glofas data, but they are stored under a different url, it is not cds, but rather ads. Switching the url in the dataset is then just simpler then going to the "config" file of the cds-api. Furthermore, with a fixed url in the data retrieval, changes can be made more easily. It was just a comment here and will be taken up in future additions.

pyproject.toml
TimFuermann: I found out that encoder=netcdf4, cannot directly be used for byte arrays, therefore h5netcdf was used.

atlite/datasets/era5.py
lkstrp: Ok, that makes sense. But I don't think we want to hardcode the url in here without the ability to pass it as an argument. We are basically overriding any urls a user may have set up manually via the cds api config file without being able to change it. So just move it to an extra argument to get this merged. Future additions we can make anyways

atlite/datasets/era5.py
lkstrp: Which makes sense. That's good enough

pyproject.toml
lkstrp: Ok. In my understanding `h5netcdf` can not be a full replacement for `netcdf4`. I'm fine with adding it as an dependency now. But leave a comment so we might wanna clean that up in future.

```suggestion
    "h5netcdf", # For new cds api. Maybe make optional / may not be needed with future netcdf4 updates
```

lkstrp:
Next to the comments, please add a release note. Afterwards we could merge this. Thank you !

atlite/datasets/era5.py
```diff
@@ -290,9 +292,9 @@ def retrieval_times(coords, static=False, monthly_requests=False):
     time = coords["time"].to_index()
     if static:
         return {
-            "year": str(time[0].year),
-            "month": str(time[0].month),
-            "day": str(time[0].day),
+            "year": [str(time[0].year)],
+            "month": [str(time[0].month).zfill(2)],
```
fneum, line 296: If leading zeros are required, perhaps this is better (and "%d", "%Y" for day and year):
```suggestion
            "month": [time[0].strftime("%m")],
```

fneum (changes requested):
Thanks @TimFuermann!

As outlined in the corresponding issue, I would first like to reproduce the problem to see if this change is necessary.

Second thing is that I would like to see a brief performance check in terms of memory and time.

maurerle:
As I read it, this PR can be closed in favor of the merged https://github.com/PyPSA/atlite/pull/454 :)

FabianHofmann:
closing as the issue is already resolved as mentioned above

PR #417, opened 2024-12-17
remove version assert

lkstrp:
- removing version assert, since people struggle with it when they not pull with tags


ekatef:
They definitely do 😄 Thanks a lot for introducing the fix!

PR #420, opened 2025-01-03
rasterio 1.4.0/1.4.1 flips y-axis in `compute_availabilitymatrix`

fneum:
This is a bit of a random find, but caused some significant faults in renewable potential calculations...

Versions 1.4.0 and 1.4.1 or `rasterio` cause a flipped y-axis in `cutout.compute_availabilitymatrix()`, which neither earlier nor later versions of `rasterio` do. With no changes but changed rasterio version:

## With `rasterio<1.4,>=1.4.2`

I think excluding these versions makes more sense than implementing version-specific behavior in `atlite`.

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -30,6 +30,9 @@ Upcoming Release
 
 * Added support for ``numpy>=2".
 
+* Exclude versions 1.4.0 and 1.4.1 of ``rasterio`` due to a bug in these
+  versions causing flipped axes in ``cutout.compute_availabilitymatrix()``.
+
 Version 0.3.0
 =============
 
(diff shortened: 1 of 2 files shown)


lkstrp (approved):
Quite a random find, great!

I think this is also what's causing the pypsa-eur problems, which is why we've pinned it to `rasterio==1.4.1` right now. Do you think this could be solved with a fix in atlite or do we need to upload updated cutouts somewhere as well? 
At least hopefully it does.

fneum:
I don't think the cutouts are affected by this.

lkstrp:
`rasterio==1.4.1` pin is no longer needed. So either the cutouts have been affected or some other rasterio dependency issue. Anyway

PR #434, opened 2025-03-27
support runoff normalization for partial years with annual normalization data

fneum:

It makes `cutout.runoff(normalize_using_yearly=...)` more lenient by supporting proportional scaling of partially represented years in the `cutout`, assuming even distribution of runoff throughout the year. In this case, a warning is printed due to the strong assumption.

Relevant for https://github.com/PyPSA/pypsa-eur/pull/1613 (e.g. July-June meteorological years, in case the original cutout does not span Jan-Dec for each year included)

This also means we can include hydro in the CI test of PyPSA-Eur, which uses only a week of weather data.

## Checklist

- ~[] Unit tests for new features were added (if applicable).~
- ~[] Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).~


atlite/convert.py
brynpickering, line 1041: `normalize_using_yearly` is super opaque as an argument. Can you type hint it and add a docstring? After more time than should be needed, I now understand it to be a `pd.Series` with a date range on the index and floats as values, but that should not have required several minutes of code inspection to work out 😅

brynpickering, line 1064: You're doing this check multiple times. I get the impression that it's because you want to leave open the option for `normalize_using_yearly` to have an arbitrary datetime index (e.g. monthly values)? My preference would be to accept only yearly values and then work on that basis from the start. This would be easiest done by formatting normalize_using_yearly to have an integer index (as datetimeindex will set the timestamp to `<year>-01-01 00:00` which could easily lead to a mistake when doing slicing.

atlite/convert.py
```diff
@@ -962,20 +962,68 @@ def runoff(
...
 
-        years = (
-            pd.Series(pd.to_datetime(result.coords["time"].values).year)
-            .value_counts()
-            .loc[lambda x: x > 8700]
-            .index.intersection(normalize_using_yearly_i)
-        )
-        assert len(years), "Need at least a full year of data (more is better)"
-        years_overlap = slice(str(min(years)), str(max(years)))
+        result_time = pd.DatetimeIndex(result.coords["time"].values)
```
brynpickering, line 965: Won't "time" _always_ be a datetime?

```suggestion
        result_time = result.coords["time"].to_index()
```

atlite/convert.py
```diff
@@ -962,20 +962,68 @@ def runoff(
...
+        overlap_start = max(result_period[0], norm_period[0])
+        overlap_end = min(result_period[1], norm_period[1])
+
+        if overlap_start >= overlap_end:
+            raise ValueError(
+                f"No overlap between runoff data ({result_period}) and normalization data ({norm_period})"
+            )
+
+        years_overlap = slice(str(overlap_start.date()), str(overlap_end.date()))
         dim = result.dims[1 - result.get_axis_num("time")]
```
brynpickering, line 989: Why expect only ever one dim? To be agnostic of this, you could do:

```suggestion
        non_time_dims = list(set(result) - {"time"}) 
```
and then assume a list later on as well (so `[dim]` -> `non_time_dims` and `reindex({dim: result.coords[dim]})` -> `reindex({dim: result.coords[dim] for dim in dims})` or `reindex_like(result.coords[dims])` (I think))

atlite/convert.py
```diff
@@ -962,20 +962,68 @@ def runoff(
...
+                    f"which may not be accurate for seasonal patterns. Consider using "
+                    f"time-based normalization data for more precise results."
+                )
+
         result *= (
-            xr.DataArray(normalize_using_yearly.loc[years_overlap].sum(), dims=[dim])
+            xr.DataArray(norm_data, dims=[dim])
             / result.sel(time=years_overlap).sum("time")
-        ).reindex(countries=result.coords["countries"])
+        ).reindex({dim: result.coords[dim]})
```
brynpickering, line 1026: Pretty sure this could just be  `result *= norm_data / result.sel(time=years_overlap).sum("time")` and you wouldn't have to also do the reindexing etc.

brynpickering (commented):
I think this could be considerably simplified if we assume `normalize_using_yearly` is always annual data. It's unclear if this is the case since the name suggests so, but the possibility of a datetime index suggests it could e.g. be monthly data.

If definitely annual, you could do something like:

```python
result_time = result.coords["time"].to_series()
annual_grouper = result_time.groupby(result_timeindex.year)
result_annual_secs = annual_grouper.apply(lambda x: (x.iloc[-1] - x.iloc[0]).total_seconds())
full_annual_secs = annual_grouper.apply(lambda x: (pd.Timestamp(str(x.name + 1)) - pd.Timestamp(str(x.name))).total_seconds())
fractions = result_annual_secs / full_annual_secs
if (sub_annual := fractions < 0.99).any(): # 99% to handle cases where the result data has a coarse frequency and therefore `result_annual_secs < full_annual_secs` even when it covers the full year
    logger.warning(...)
overlap = normalize_using_yearly.index.intersection(fractions.index) # assuming you've previously normalised the index to integer years
if not overlap:
    raise ValueError("no overlap ...")
overlap_slice = slice(overlap.min(), overlap.max()) # for use later

norm_data = normalize_using_yearly.mul(fractions, fill_value=0)
```

Even if _not_ definitely annual, the overall amount of code in this conditional can be significantly reduced to make it easier to maintain (I just haven't considered exactly what would be needed)

PR #445, opened 2025-07-02
Add argument logic for the arguments into cutout init and prepare

GohNgeeJuay:
Closes # (if applicable).

## Checklist


GohNgeeJuay:
Not sure why that previous commit is in this PR, but I need to remove that :)

atlite/cutout.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, path, **cutoutparams):
             Time range to include in the cutout, e.g. "2011" or
             ("2011-01-05", "2011-01-25")
             This is necessary when building a new cutout.
-        bounds : GeoSeries.bounds | DataFrame, optional
+        bounds : DataFrame | list, optional
```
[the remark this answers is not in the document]
GohNgeeJuay: Also, I am not sure what arguments that we are allowing, so please correct me!

atlite/data.py
```diff
@@ -199,6 +206,10 @@ def cutout_prepare(
         logger.info("Cutout already prepared.")
         return cutout
 
+    # ensure that the tmpdir actually exists
+    temp_dir_path = Path(tmpdir)
+    if not temp_dir_path.is_dir():
+        raise NotADirectoryError(f"No directory was found at {temp_dir_path}")
```
GohNgeeJuay, line 212: I realized this silly mistake when I gave a tmpdir which does not exists because I have not created it yet. The confusing part was the stack trace which seems to be pointing into some deep functions, so I thought there was something else that cause the error (but it was simply because my tmpdir did not exist). 

image

So I think it would be nice if we fail fast here and give a clear indication that the directory does not exist. But a point can be made about putting this into the relevant function that creates the temporary directory. Not sure what is best.

euronion: We could either fail nicely as you suggest, or just create the directory silently.

Note that `NotADirectoryError` is the wrong error to use in this case (https://docs.python.org/3/library/exceptions.html#NotADirectoryError). The correct one would be `FileNotFoundError`, which also applies to directories.

GohNgeeJuay:
On the other note, I totally get that we do not want to simply change things which can affect dependencies elsewhere. This is just something that I thought would help people who are just trying things out for the first time.

I am quite interested in contributing to this project in my free time. If you have any more important tasks/issue which you think is suitable for people like me (who are new to the project), please let me know!

atlite/cutout.py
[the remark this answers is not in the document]
GohNgeeJuay: Yes, its a tuple. I corrected the type hints. 

The `DataFrame` as argument hasn't been working for me, so I added a logic for that below.

atlite/data.py
GohNgeeJuay: Alright, I have changed the error type.

I think creating the directory silently would also work, but it brings up the question on how strict the code should be in handling user inputs. I think we can be strict on this so that users can recheck their inputs to prevent creation of files in unintended paths. But let me know what you think!

atlite/cutout.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, path, **cutoutparams):
             Time range to include in the cutout, e.g. "2011" or
             ("2011-01-05", "2011-01-25")
             This is necessary when building a new cutout.
-        bounds : GeoSeries.bounds | DataFrame, optional
+        bounds : GeoSeries.bounds | DataFrame | Tuple, optional
             The outer bounds of the cutout or as a DataFrame
             containing (min.long, min.lat, max.long, max.lat).
```
euronion, line 103:
```suggestion
        bounds : DataFrame | Tuple, optional
            The outer bounds of the cutout containing (min.long, min.lat, max.long, max.lat)
            or a single-row DataFrame with [["minx", "miny", "maxx", "maxy"]] column values.
            From GeoPandas DataFrames and Series, this is easily accessible through `.geometry.bounds`.
```

I checked their documentation again and noticed that `bounds` are not a class, only a tuple attribute. I've moved it into the docstring.

atlite/cutout.py
```diff
@@ -163,8 +163,15 @@ def __init__(self, path, **cutoutparams):
...
             if "bounds" in cutoutparams:
-                x1, y1, x2, y2 = cutoutparams.pop("bounds")
-                cutoutparams.update(x=slice(x1, x2), y=slice(y1, y2))
+                bounds = cutoutparams.pop("bounds")
+                # If its a dataframe, we will extract the values
+                if isinstance(bounds, pd.DataFrame):
+                    row = bounds.iloc[0]
+                    x1, y1, x2, y2 = row[["minx", "miny", "maxx", "maxy"]].to_list()
+                else:  # If its a list
+                    x1, y1, x2, y2 = bounds
```
euronion, line 172: Your suggestion would always take the first row, even if the user falsely provides a multi-row dataset.
Instead we should raise an error for the user if the DataFrame has != 1 row.

How about:

```suggestion
                # If its a dataframe, we will extract the values
                if isinstance(bounds, pd.DataFrame) and bounds.shape[0] == 1:
                    x1, y1, x2, y2 = bounds.iloc[0][["minx", "miny", "maxx", "maxy"]].to_list()
                elif isinstance(bounds, tuple):  # If its a list
                    x1, y1, x2, y2 = bounds
                else:
                    raise ValueError("`bounds` must be a tuple or a DataFrame with a single row entry.")
```

euronion (changes requested):
Thanks for clarifying the intention. This makes it much clearer!

With that in mind, I looked through the PR again.
Some of the code parts haven't been touched in years, so it is good that you spotted these things and helped to clean them up!

I've left some more comments to align how the functions behave with the rest.
E.g. the error message raised.

For the folder: Let's not create it, but instead fail and have the user fix it - like you suggested!

GohNgeeJuay:
I have done the suggested changes and cleaned up all of the commits.

lkstrp (approved):

Issue #442, opened 2025-05-12
cds api on macos ci fails because of docs update


lkstrp:
### Version Checks (indicate both or one)

- [ ] I have confirmed this bug exists on the lastest [release](https://github.com/pypsa/atlite/releases) of atlite.

- [x] I have confirmed this bug exists on the current [`master`](https://github.com/pypsa/atlite/tree/master) branch of atlite.

### Issue Description

#433 still leads to issues and was reverted

PR #456, opened 2025-07-23
 ci: use environment variables for CDS API configuration

lkstrp:
Maybe this fixes the miracle in #442

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -115,6 +115,7 @@ jobs:
         return 0 # Ignore windows and pytest 3.13
 
     - name: Upload code coverage report
+      if: github.event_name != 'schedule'
       uses: codecov/codecov-action@v5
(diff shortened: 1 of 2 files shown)


euronion:
You don't need to make modifications to the `era5.py` module - the `cdsapi` reads the environment variables by default (I just checked it locally).
Rather, I'd remove the creation of the `.cdsapirc` file in the tests and only go with the env variables to avoid having two parallel systems of providing the credentials.

euronion:
> > Rather, I'd remove the creation of the `.cdsapirc` file in the tests and only go with the env variables to avoid having two parallel systems of providing the credentials.
> 
> But this is what I do, isn't it? Have I missed a hardcoded `.cdsapirc` somewhere?

`*head scratching*` I could've sworn that this change was not in there when I viewed it. LGTM!

PR #475, opened 2025-11-05
wave energy converters and new wind and wave data modules

lmezilis:
Closes # (if applicable).

## Checklist


lmezilis:
Thank you!

I think we can review it right away, and I can prepare some files for documentation. Apologies for being very new to github procedures. I am slowly starting to get the hang of it.

atlite/datasets/mrel_wave.py
```diff
@@ -0,0 +1,108 @@
...
+
+    return ds
+
+
+def get_data_wave_period(ds):
+    ds = ds.rename({"tp": "wave_period"})
+    # ds["wave_period"] = (1 / ds["wave_period"])
+    ds["wave_period"] = ds["wave_period"].clip(min=0.0)
+
+    return ds
```
brynpickering: Keep the process the same as in `era5.py` - split this into data retrieval and then sanitisation and call the sanitisation function only if requested in `get_data` (you can copy most of the functionality directly over from `era5.py`)

atlite/datasets/mrel_wave.py
```diff
@@ -0,0 +1,108 @@
...
+        logger.error('Argument "data_path" not defined')
+        raise ValueError('Argument "data_path" not defined')
+    path = creation_parameters["data_path"]
+
+    ds = xr.open_dataset(path)
+
+    if "longitude" in ds and "latitude" in ds:
+        ds = ds.rename({"longitude": "x", "latitude": "y"})
+
+    ds = ds.sel(x=as_slice(cutout.extent[:2]), y=as_slice(cutout.extent[2:]))
```
brynpickering:
```suggestion
    bounds = _bounds(cutout.coords, pad=creation_parameters.get("pad", 0))
    ds = ds.sel(**bounds)
```

atlite/resources/wecgenerator/Farshore_750kW.yaml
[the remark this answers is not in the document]
brynpickering: Same for the other YAML files.

atlite/convert.py
```diff
@@ -653,6 +654,67 @@ def wind(
...
+    da = da.fillna(0)
+
+    return da
+
+
+def wave(cutout, wec_type, **params):
+    """
+    Generate wave generation time series
+
+    evaluates the significant wave height (Hs) and wave peak period (Tp)
```
brynpickering: Since `Hs` and `Tp` are MREL-specific and wouldn't make sense if using ERA5 data, it might make more sense to not reference them by these acronyms in this file.

.gitignore
```diff
@@ -27,3 +27,4 @@ paper
 # Ignore IDE project files
 .idea/
 .vscode
+.vs
```
brynpickering: It's generally best to add these pointers to your own ["global" gitignore](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/), rather than to every project you work on. That way, it never accidentally slips in without you realising it!

brynpickering (changes requested):
Thanks for opening this PR @lmezilis ! As @euronion has mentioned, some documentation is needed. My comments are about implementation. I have no comments on the method, I'm sure you've got that right! There's lots of scope to clean things up, though, which will help us maintain the feature in future. 

If you want any clarification on comments/suggestions I've made, just reply directly on the comment. If you are happy with a suggestion, you can just accept it and it will automatically update the code for you. You can always accept a suggestion but then make edits to it later.

One thing missing from the new files is a REUSE header. You can find equivalents in other files. The easiest is to just copy the one across from `era5.py`. Then, make sure your name is in `AUTHORS.rst` so you count as one of the "Contributors to atlite".

atlite/datasets/mrel_wave.py
```diff
@@ -0,0 +1,108 @@
...
+
+features = {"wave_height": ["wave_height"], "wave_period": ["wave_period"]}
+
+
+def _rename_and_clean_coords(ds):
+    """
+    Rename 'longitude' and 'latitude' columns to 'x' and 'y' and fix roundings.
+
+    Optionally (add_lon_lat, default:True) preserves latitude and
+    longitude columns as 'lat' and 'lon'.
```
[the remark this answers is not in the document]
lmezilis: this option was not necessary for the mrel wave files. I eventually renamed the dimentions in the last part of the script. I would say that the function "_rename_and_clean_coords" can be deleted here. do you think we should include this option?

EDIT: I saw the rest of the comments now, I will include the function and try to make it as similar to era5 as possible.

atlite/datasets/mrel_wave.py
```diff
@@ -0,0 +1,108 @@
...
+    #     ds = regrid(ds, coords["x"], coords["y"], resampling=Resampling.average)
+
+    logger.info("Obtaining wave data.")
+
+    ds = get_data_wave_height(ds)
+    ds = get_data_wave_period(ds)
+
+    # ds = ds.assign_coords(x=ds.coords["x"], y=ds.coords["y"])
+
+    return ds
```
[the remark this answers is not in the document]
lmezilis: Thank you very much for this suggestion.

brynpickering (commented):
Thanks for the changes @lmezilis. Just a couple of extra comments.

I still need to check the processing method, which I'll do now. In the meantime, it would be great to add some documentation. The easiest would be to add a jupyter notebook to `examples` and then link that into the docs by adding an entry into `doc/examples` (following the example of the other entries). The example notebook could load from era5 and mrel separately and maybe compare the results for a specific gridcell?

euronion:
> @euronion not sure what the system should be for accessing the data. This PR works on the basis that the user downloads the data themselves. The data is available via OpenDAP so it would be feasible to query them directly with the appropriate lat/lon/time attrs, e.g.:
> 
> ```python
> import xarray as xr
> years = [2000]
> months=[1, 2, 3, 4, 5, 6]
> remote_data = xr.open_mfdataset(
>     [f"https://opendap.4tu.nl/thredds/dodsC/data2/djht/f359cd0f-d135-416c-9118-e79dccba57b9/1/{year}/TU-MREL_EU_ATL-2M_{year}{month:02}.nc?hs,latitude,longitude" for year in years for month in months],
>     engine="netcdf4"
> )
> # `time` coord seems to be corrupted in the data source, so we have to translate integers to datetime locally
> remote_data.coords["time"] = pd.date_range(f"{years[0]}-{months[0]:02}", f"{years[-1]}-{months[-1]:02}", freq="1H", inclusive="left")
> remote_data.sel(latitude=y, longitude=x, time=time)
> ```

This data get's read into a `Cutout`, right?

In an ideal world we support both:
* Automatically retrieving of the data
* Building from a local downloaded file

If it is easily possible, then yes, please implement automatic retrieval as well.
Since you already have building from a local file, you could implement it by downloading to a local temporary file and then pass that to your existing function.

One question: The naming "wecgenerator" strikes me a bit odd. I haven't looked at it in detail, although my understanding is that in this PR the conversion is from wave energy to electricity. Wouldn't then the term "wec" or "waveenergyconverter" be more appropriate, as not only the "generator", but the whole system is modelled?

lmezilis:
>  the variable `tp` doesn't exist; only `t01` and `t02` exist linked to wave period data (both being mean wave periods). How did you get `tp`? From a different version of this dataset?

@brynpickering I must have changed the variables to be more convenient to work back in the day and completely forgot their original names. `t01` is the one that should be used. That dataset has a lot of different types of ocean variables.  


> In the meantime, it would be great to add some documentation.

of course, sorry for the delays, I am working on this, I have my documentation ready but trying to make it apealing and clear. probably be ready by tomorrow

> compare the results for a specific gridcell?

I assume between ERA5 and ECHOWAVE, correct?

lmezilis:
> One question: The naming "wecgenerator" strikes me a bit odd.

@euronion you are right, a simple `wec` should do it. I will update this. Also in pypsa-eur I used the term `wec_type` as a generator index in `build_renewables`. I notice that `wind turbines` are called `turbines`. Should I just type `wec` there too?

brynpickering:
@lmezilis no worries. We'll probably squash all these commits when we merge it in, so it'll all be cleaned up.

You could install pre-commit locally so the fixes are managed locally. In your `atlite` working environment call `pre-commit install`. Then pre-commit will fix things before you try to commit.

RE allowing data downloads, I've found that the OpenDAP fails when trying to download more than a few MB of data at once (`DAP failure` or `Authorization failure`). Not sure if you get this issue @lmezilis but it seems to me that it's too volatile to rely upon as a way to access the data.

FabianHofmann:
@lmezilis we have some merge conflicts now. Do you want to resolve them by yourself or should I take over? I would review afterwards

lmezilis:
@FabianHofmann I made some adjustments and finally the pre-commit check is completed. I think we are ready to finalize this PR.

PR #492, opened 2026-03-16
Enforce strict mypy type checking and additional ruff rules

FabianHofmann:

This PR enforces strict static type checking with mypy and enables additional ruff lint rules across the entire codebase (including `test/` and `doc/`). It also fixes one bug uncovered while annotating `line_rating`.

### Bug fixes

- **`Cutout.line_rating`**: the wrapper computed line azimuth in radians and passed it to `convert_line_rating`, which interpreted `psi` as degrees and did `radians(psi)` internally — effectively dividing the angle by ~57. Azimuths are now computed in degrees (`line_azimuth_degrees`), matching the documented unit. Added a focused parametrized test and a release note. See [line_rating_azimuth_bugfix.html](https://github.com/user-attachments/files/27955165/line_rating_azimuth_bugfix.html) for more details

### Docstrings

- Add docstrings to all functions

### Mypy

- Configured strict mypy settings in `pyproject.toml`: `disallow_incomplete_defs = true`, `check_untyped_defs = true`
- Removed all directory exclusions from mypy — it now checks `atlite/`, `test/`, and `doc/`
- Updated `.pre-commit-config.yaml` to match
- Added `atlite/_types.py` with a small set of `TypeAlias` definitions (kept lean per review: `NDArray`, `PathLike`, `CrsLike`, `ConvertResult`, plus `Literal` aliases for tracking/clearsky/orientation/etc.)
- Added missing type annotations to function signatures (e.g. conversion functions in `convert.py`)
- Fixed `__all__` in `__init__.py` to use strings instead of object references

### Ruff

Enabled 9 new rule sets and fixed all violations (~260 fixes):

- **B** (flake8-bugbear): Added `from err`/`from None` to re-raised exceptions, fixed mutable defaults
- **SIM** (flake8-simplify): Collapsed nested `if` statements, simplified dict operations
- **RET** (flake8-return): Removed superfluous `else` after `return`/`raise`
- **C4** (flake8-comprehensions): Simplified unnecessary `dict()`/`list()` constructors
- **TC** (flake8-type-checking): Moved type-only imports into `TYPE_CHECKING` blocks
- **NPY** (numpy): Replaced legacy `np.random.seed`/`np.random.rand` with `np.random.default_rng`
- **G** (flake8-logging-format): Replaced f-strings in logging with lazy `%` formatting
- **PTH** (flake8-use-pathlib): Replaced `os.path.*` and `open()` with `pathlib` equivalents
- **RUF100**: Cleaned up unused `# noqa` comments

### Review follow-ups

- Restored non-obvious comments, citations, and docstring context flagged by review
- Dropped `ERA5RetrievalParams` and trimmed `_types.py` to remove over-abstracted aliases
- Removed redundant local variable annotations in `data.py`
- Simplified `regrid()` in `gis.py` (dropped wide `cast()` that broke chain formatting)

## Checklist


atlite/datasets/__init__.py
```diff
@@ -2,10 +2,15 @@
...
+
+from typing import TYPE_CHECKING
 
 from atlite.datasets import era5, gebco, sarah
 
-modules = {"era5": era5, "sarah": sarah, "gebco": gebco}
+if TYPE_CHECKING:
+    from types import ModuleType
+
+modules: dict[str, ModuleType] = {"era5": era5, "sarah": sarah, "gebco": gebco}
```
lkstrp, line 16: Why is ModuleType needed here?

atlite/datasets/cordex.py
```diff
@@ -29,38 +38,111 @@
...
+    tuple of ((int, int), xr.Dataset)
+        ``(year, month)`` key and the corresponding monthly dataset slice.
+    """
     with xr.open_dataset(fn) as ds:
         ds = rename_and_clean_coords(ds)
         ds = ds.rename({oldname: newname})
         ds = ds.sel(x=xs, y=ys)
 
         if newname in {"influx", "outflux"}:
-            # shift averaged data to beginning of bin
```
lkstrp, line 48: Why remove the comments?

atlite/datasets/cordex.py
```diff
@@ -29,38 +38,111 @@
...
             ds = ds.assign_coords(
                 time=(
                     pd.to_datetime(ds.coords["time"].values) - pd.Timedelta(hours=1.5)
                 )
             )
         elif newname in {"runoff"}:
-            # shift and fill 6hr average data to beginning of 3hr bins
```
lkstrp, line 55: Why remove the comments?

atlite/datasets/cordex.py
```diff
@@ -115,138 +295,150 @@ def tasks_yearly_cordex(
...
-    "temperature": dict(
-        tasks_func=tasks_yearly_cordex,
-        prepare_func=prepare_data_cordex,
-        oldname="tas",
-        newname="temperature",
-        template=os.path.join(
-            config.cordex_dir,  # noqa: F821
+weather_data_config: dict[str, dict[str, Any]] = {}
+try:
+    from atlite import config  # type: ignore[attr-defined]
```
lkstrp, line 314: That should never be an import error is it?

atlite/datasets/cordex.py
```diff
@@ -115,138 +295,150 @@ def tasks_yearly_cordex(
...
+                "altitude",
+                "orog_*.nc",
+            ),
+        },
+        "CWT": {
+            "tasks_func": tasks_yearly_cordex,
+            "prepare_func": prepare_weather_types_cordex,
+            "oldname": "CWT",
+            "newname": "CWT",
+            "template": os.path.join(  # noqa: PTH118
```
lkstrp, line 418: I would rather ignore PTH118 globally for now, so we can add it easily later again. Using pathlib instead of os is very recommended

atlite/datasets/cordex.py
```diff
@@ -115,138 +295,150 @@ def tasks_yearly_cordex(
...
+                "CWT_*_{year}*.nc",
+            ),
+        },
+    }
+except ImportError:
+    pass
+
+meta_data_config: dict[str, Any] = {}
+try:
+    from atlite import config  # type: ignore[attr-defined]
```
lkstrp, line 431: Same question for the import error

atlite/datasets/era5.py
```diff
@@ -62,48 +71,73 @@ def nullcontext():
...
-def _add_height(ds):
+def _add_height(ds: xr.Dataset) -> xr.Dataset:
     """
-    Convert geopotential 'z' to geopotential height following [1].
+    Convert geopotential to height and replace the 'z' variable.
 
-    References
+    Parameters
     ----------
-    [1] ERA5: surface elevation and orography, retrieved: 10.02.2019
```
lkstrp, line 71: Reference section was removed here

atlite/datasets/era5.py
```diff
@@ -62,48 +71,73 @@ def nullcontext():
...
 
-    Optionally (add_lon_lat, default:True) preserves latitude and
-    longitude columns as 'lat' and 'lon'.
+    Returns
+    -------
+    xr.Dataset
+        Dataset with standardized coordinates.
     """
     ds = ds.rename({"longitude": "x", "latitude": "y", "valid_time": "time"})
-    # round coords since cds coords are float32 which would lead to mismatches
```
lkstrp, line 92: Why remove it?

atlite/datasets/era5.py
```diff
@@ -125,27 +159,48 @@ def get_data_wind(retrieval_params):
         np.log(ds["wnd10m"] / ds["wnd100m"]) / np.log(10 / 100)
     ).assign_attrs(units="", long_name="wind shear exponent")
 
-    # span the whole circle: 0 is north, π/2 is east, -π is south, 3π/2 is west
```
lkstrp, line 128: Why remove the comment?

atlite/datasets/era5.py
```diff
@@ -170,38 +225,53 @@ def get_data_influx(retrieval_params):
     )
     ds = ds.drop_vars(["ssrd", "ssr"])
 
-    # Convert from energy to power J m**-2 -> W m**-2 and clip negative fluxes
     for a in ("influx_direct", "influx_diffuse", "influx_toa"):
         ds[a] = ds[a] / (60.0 * 60.0)
         ds[a].attrs["units"] = "W m**-2"
 
-    # ERA5 variables are mean values for previous hour, i.e. 13:01 to 14:00 are labelled as "14:00"
```
lkstrp, line 178: Why remove the comments?

atlite/datasets/era5.py
```diff
@@ -213,79 +283,115 @@ def get_data_temperature(retrieval_params):
...
+    -------
+    xr.Dataset
+        Dataset with non-negative runoff values.
     """
     ds["runoff"] = ds["runoff"].clip(min=0.0)
     return ds
 
 
-def get_data_height(retrieval_params):
+def get_data_height(retrieval_params: ERA5RetrievalParams) -> xr.Dataset:
```
lkstrp, line 331: See comment in `_types.py`

atlite/datasets/era5.py
```diff
@@ -350,40 +491,35 @@ def sanitize_chunks(chunks, **dim_mapping):
...
-    grib_file: str | Path, chunks=None, tmpdir: str | Path | None = None
+    grib_file: PathLike,
+    chunks: dict[str, int] | None = None,
+    tmpdir: PathLike | None = None,
 ) -> xr.Dataset:
     """
-    Convert grib file of ERA5 data from the CDS to netcdf file.
+    Open a GRIB file using cfgrib with standardized coordinate conventions.
 
-    The function does the same thing as the CDS backend does, but locally.
```
lkstrp, line 358: Do we remove this on purpose?

atlite/datasets/era5.py
```diff
@@ -397,10 +533,6 @@ def open_with_grib_conventions(
         add_finalizer(ds, grib_file)
 
     def safely_expand_dims(dataset: xr.Dataset, expand_dims: list[str]) -> xr.Dataset:
-        """
-        Expand dimensions in an xarray dataset, ensuring that the new dimensions are not already in the dataset
-        and that the order of dimensions is preserved.
```
lkstrp, line 402: Removed on purpose?

atlite/datasets/era5.py
```diff
@@ -432,140 +562,122 @@ def safely_expand_dims(dataset: xr.Dataset, expand_dims: list[str]) -> xr.Datase
...
     lock: SerializableLock | None = None,
-    **updates,
+    **updates: Any,
 ) -> xr.Dataset:
     """
-    Download data like ERA5 from the Climate Data Store (CDS).
-
-    If you want to track the state of your request go to
-    https://cds-beta.climate.copernicus.eu/requests?tab=all
+    Download ERA5 data from the CDS API and return as an xarray Dataset.
```
lkstrp, line 570: Also a lot of quite useful stuff is removed (link to cds platform and Example)

atlite/pv/irradiation.py
```diff
@@ -2,20 +2,61 @@
...
+    )
+
 logger = logging.getLogger(__name__)
 
 
-def DiffuseHorizontalIrrad(ds, solar_position, clearsky_model, influx):
-    # Clearsky model from Reindl 1990 to split downward radiation into direct
-    # and diffuse contributions. Should switch to more up-to-date model, f.ex.
-    # Ridley et al. (2010) http://dx.doi.org/10.1016/j.renene.2009.07.018 ,
-    # Lauret et al. (2013):http://dx.doi.org/10.1016/j.renene.2012.01.049
```
lkstrp, line 17: All references are removed

atlite/pv/irradiation.py
```diff
@@ -41,8 +76,6 @@ def DiffuseHorizontalIrrad(ds, solar_position, clearsky_model, influx):
             + (k >= 0.78) * fmax(0.1, 0.486 * k - 0.182 * sinaltitude)
         )
     elif clearsky_model == "enhanced":
-        # Enhanced Reindl model with ambient air temperature and relative
-        # humidity
```
lkstrp, line 45: Why remove the comment?

atlite/pv/irradiation.py
```diff
@@ -85,13 +139,9 @@ def TiltedDiffuseIrrad(ds, solar_position, surface_orientation, direct, diffuse)
     influx = direct + diffuse
 
     with np.errstate(divide="ignore", invalid="ignore"):
-        # brightening factor
```
lkstrp, line 88: Why remove comments?

atlite/_types.py
```diff
@@ -0,0 +1,54 @@
...
+import scipy.sparse as sp
+import xarray as xr
+from pyproj import CRS
+from shapely.geometry.base import BaseGeometry
+
+NDArray: TypeAlias = np.ndarray[Any, np.dtype[np.floating[Any]]]
+NDArrayInt: TypeAlias = np.ndarray[Any, np.dtype[np.signedinteger[Any]]]
+NDArrayBool: TypeAlias = np.ndarray[Any, np.dtype[np.bool_]]
+DataArray: TypeAlias = xr.DataArray
+Dataset: TypeAlias = xr.Dataset
```
lkstrp, line 21: Do we need all those? Some are just aliases. I think it's much cleaner to just keep for example use `xr.Dataset` than redefining it?

atlite/_types.py
```diff
@@ -0,0 +1,54 @@
...
+NumericArray: TypeAlias = NDArray | DataArray
+Number: TypeAlias = int | float | np.number[Any]
+GeoDataFrame: TypeAlias = gpd.GeoDataFrame
+GeoSeries: TypeAlias = gpd.GeoSeries
+Geometry: TypeAlias = BaseGeometry
+CrsLike: TypeAlias = str | int | CRS | dict[str, Any] | None
+SparseMatrix: TypeAlias = sp.lil_matrix | sp.csr_matrix
+
+TrackingType: TypeAlias = (
+    Literal["horizontal", "tilted_horizontal", "vertical", "dual"] | None
```
lkstrp, line 32: I like to define those literals! But I would not put `| None` in there in general. That information if something is optional should be in the signature. And again, I am not so sure if this needs to be defined in an extra Type? How often is this actually used in the signature? If I wanna see this in the function I need to click once more now. Guess for some it makes sense, but you think for all of them?

atlite/_types.py
```diff
@@ -0,0 +1,54 @@
...
+)
+ClearskyModel: TypeAlias = Literal["simple", "enhanced"]
+TrigonModel: TypeAlias = Literal["simple", "perez"]
+IrradiationType: TypeAlias = Literal["total", "direct", "diffuse", "ground"]
+HeatPumpSource: TypeAlias = Literal["air", "soil"]
+OrientationName: TypeAlias = Literal["latitude_optimal", "constant", "latitude"]
+DataFormat: TypeAlias = Literal["grib", "netcdf"]
+
+
+class ERA5RetrievalParams(TypedDict, total=False):
```
lkstrp, line 42: Is it wise to hardcode that here? Since it's a dict it doesn't give you any benefits, is it?

lkstrp (changes requested):
Tons of good cleanup changes, thank youuu ! Just a bit too much and mixed which makes it a bit hard to review, since this now includes changed docstrings, ruff, types and tests.

I think some things need another look. I only went through parts for it now:
- There are many comments removed, and sometimes it looks like Claude chucked them, but some might be removed intentionally? I am not always sure and I would wanna keep hand crafted code explaining comments in general (if useful and up to date). The same goes for docstrings sometimes. They are rewritten, often improved, but sometimes useful old information is removed
- In `_types.py`, many types are defined which are then used everywhere. This makes sense for new literals and more complicated types, but some also just define an alias for existing types from the library. I suppose this is also a style question, but since I believe that simplicity is becoming increasingly important, I would try to simplify it. If an argument requires many different types, this is also an indication that the signature should be restructured. Abstracting all those types away just hides that

atlite/datasets/cordex.py
FabianHofmann: I just don't like comments which are describing the code

atlite/datasets/cordex.py
FabianHofmann: looking at this closer. this is a bug in master as well. the `config` namespace was introduced in https://github.com/PyPSA/atlite/pull/388 and is not defined. guess it hasn't come up as a but as nobody is using the cordex data anymore

atlite/datasets/era5.py
FabianHofmann: reverted

atlite/_types.py
FabianHofmann: removed

FabianHofmann:
@lkstrp I reverted mostly all removals of comments and made the changes to `_types.py` leaner. also discovered a quite important bug that is fixed now here as well. I guess you can take another look now

lkstrp (approved):

PR #504, opened 2026-04-26
Update CSP

awongel:

CSP conversion now models the actual collector geometry instead of assuming perfect sun-tracking. Both technologies are computed via `SurfaceOrientation` with `TiltedDirectIrrad = (cosincidence / sin(altitude)) * influx_direct`. Solar tower (`tracking="dual"`) yields `cosincidence = 1`, so its output is unchanged DNI. Parabolic trough now uses `tracking="horizontal"` (single-axis tracking), so it collects less than DNI when the sun is not aligned with its rotation axis — correcting the previous overestimate. The old `atlite/csp.py` module (with its hardcoded 3.75° altitude clamp) was removed and folded into `convert.py`; the public `cutout.csp()` API is unchanged. Reworked the `working-with-csp` example to showcase the difference.

## Checklist


atlite/gis.py
```diff
@@ -855,7 +853,8 @@ def _reproject(src, **kwargs):
             output_core_dims=[["yout", "xout"]],
             output_dtypes=[dtypes.pop()],
             dask_gufunc_kwargs=dict(
-                output_sizes={"yout": dst_shape[0], "xout": dst_shape[1]}
+                output_sizes={"yout": dst_shape[0], "xout": dst_shape[1]},
+                allow_rechunk=True,
```
fneum: Why was this change necessary?

fneum (changes requested):
Thanks @awongel! Looks great. I have two suggestions / requests:

1. Could you remove the Saudia Arabia example? The example got quite long and to me it didn't add too many new insights.

2. Do you have a comparison of the `convert_csv()` output before and after? That would make it easier to see whether the changes are plausible. Just a  graphic + code snippet would be helpful.

atlite/gis.py
awongel: Hi @fneum!

1. I had discussed with @euronion what we should add and felt that a second location would be useful to show that it works more generally than in a single location. If you think it's too much, I don't mind removing it again.

2. Since I replaced the calculation, would you suggest to keep the previous way of calculating it somewhere in the code? I wanted to avoid that. Or save the old result somewhere, but then you can't immediately see how it was calculated.

The change in the gis.py file was needed for the sarah data part of the example, which wouldn't run for me otherwise. Happy to remove if you can confirm this is not needed.

FabianHofmann:
@awongel should I update the code according to @fneum review? I would also argue we should remove the `allow_rechunk` for now and generalize the exposure of dask kwargs.

FabianHofmann:
@awongel took the freedom to merge the latest master and remove the saudi arabia example. we should keep the allow_rechunk in as it is generally needed for sarah cutouts and regrid operations. I have the changes locally but don't want to push on your master unless you say its okay. alternatively I can raise another pr superseding this one

awongel:
@FabianHofmann, awesome, thanks for picking this up again! Yes, please feel free to push your changes directly to my master.
Absolutely, will follow your judgement on `allow_rechunk`. (It just felt wrong to push something that wouldn't run for me.)

FabianHofmann (approved):
thanks @awongel for the contribution. I updated the release notes and want to update the pr summary for the record. (would be nice in future to have a summary written by the author)

awongel:
Thanks @FabianHofmann! Happy to write a summary if you point me to what you mean, I thought the description at the beginning was the summary.

FabianHofmann:
> Thanks @FabianHofmann! Happy to write a summary if you point me to what you mean, I thought the description at the beginning was the summary.

Exactly, but note the current summary is a what I wrote (GitHub allows editing comments from others), before it was empty. So, no worries

PR #507, opened 2026-05-11
Add alternative source for ERA5 data to eliminate queueing

mrmorawski:
Closes #495

Currently the ERA5 dataset is downloaded from CDS. This requires authentication and users often get stuck in a processing queue for hours.

This PR adds support for downloading ERA5 from a mirror hosted at NCAR via OPeNDAP. This mirror requires no authentication and there is no queue. 

The modification is implemented as a new dataset module called `era5-ncar`, which exposes the same API as the current `era5.py` module. All the user needs to do is change one line when creating the cutout. Code remains fully backwards-compatible.

```python
cutout = atlite.Cutout(
    path="spain-2024.nc",
    module="era5-ncar",    # <-- only change from "era5"
    x=slice(-10, 5),
    y=slice(35, 44),
    time="2013-01-01
   )
```

### Comparison of `era5` vs `era5-ncar`
The NCAR mirror is stored as NetCDF files at the default 0.25/0.25 deg. resolution used by CDS. This has implications for performance and accuracy.

#### Performance
Tested on a Germany-sized cutout, 6 months of data, downloaded at the default grid size (CDS files served from CDS server cache, no queuing):

| Source       | Wall time | Peak memory usage | received files |
| ------------ | --------: | ----------------: | -------------: |
| NCAR THREDDS |   239.3 s |            865 MB |       447.6 MB |
| CDS          |   194.2 s |            802 MB |       186.4 MB |

As you can see, we need to download slightly more data from NCAR and do more local processing, but the extra time is typically offset by the CDS queue time (over an hour when this cutout was first requested). Coarser/larger cutouts will benefit CDS, while smaller cutouts will have an even larger speedup on NCAR.

#### Accuracy
Results on the default grid (and its multiples) are very close, within float32 precision. Results on other grids are slightly off, as we cannot reproduce the exact interpolation logic used by CDS. Details of the interpolation errors are explained in the example notebook `examples/comparing_era5_sources_cds_vs_ncar.ipynb`

#### When to use which source
I included the following line in the documentation:

> **So which source should I use?** For very large cutouts (e.g. continents larger than Europe), for workflows that already use CDS, and when you require data at resolution that is not a multiple of the default 0.25/0.25 deg. grid, use CDS. For everything else, NCAR is more convenient.

## Communication with UCAR
I asked UCAR, who administer this dataset for NCAR, whether they were OK with the OPeNDAP access. The response was positive. If we experience rate limiting on the server side, we may need to limit the number of simultaneous connections further (I capped it at 8).

## LLM use disclosure
I used LLMs to:
- help me understand the THREDDS API
- write some of the processing code
- review the module
- write the benchmarking and plotting code

I manually wrote all the documentation and most of the docstrings. I hand-checked every line of code in this PR.

## Remaining tests
I'd appreciate if someone could test the module on their machine on a large download. I've been test-running it for a week on my own setup, and it was stable (not a single failed download). However, there may still be hidden networking bugs.

## Other ERA5 sources
I investigated other sources for ERA5. This is, to my knowledge, the only one fulfills the following criteria:
- ran by a large organisation
- easy spatial subsetting
- no authentication and queueing
- provides full set of the variables used by `atlite`

The ARCO datasets stored by Google, Microsoft and Amazon seem to be chunked along the time dimension only, so one always has to download a whole world's worth of data.

Earthmover stores era5 on s3 in Zarr, chunked in yearly, 12x12 spatial chunks, but they only do surface variables. 

## Checklist

- [x] ~~Newly introduced dependencies are added to `environment.yaml`, `environment_docs.yaml` and `setup.py` (if applicable).~~ I added dependencies to `pyproject.toml`, I don't see an `environment.yaml`


fneum:
Incidentally, I stumbled across another alternative yesterday: The Earth Data Hub.

https://earthdatahub.destine.eu/collections/era5/datasets/reanalysis-era5-single-levels

```python
import xarray as xr

PAT = "your-personalised access token" # https://earthdatahub.destine.eu/getting-started
ds = xr.open_dataset(
    f"https://edh:{PAT}@data.earthdatahub.destine.eu/era5/reanalysis-era5-single-levels-v0.zarr",
    chunks={},
    engine="zarr",
)

cutout_pre = ds[["u100", "v100"]].sel(
    latitude=slice(50, 47), longitude=slice(5, 8), valid_time="2025-10-01"
)

cutout = cutout_pre.compute()  # triggers download

cutout.to_netcdf("cutout.nc")
```

I could download fully European dataset in 3 minutes we'd normally use in PyPSA-Eur and the access seems quite elegant (there's also a `.netrc` file-based auth option).

This seems to fulfil all your criteria except authentication. What do you think? @mrmorawski 

Maybe it's even close enough to CDSAPI output so that it could be used within `era5.py` module.

mrmorawski:
That does look like a really good option @fneum !

Zarr is a easier to work with than DAP2. EDH chunks are `(4320, 64, 64)` (half year, ~2000x2000km slice), which would be a great fit for `atlite`. Plus, the future of NCAR is uncertain [in the current political climate](https://www.nature.com/articles/d41586-026-01501-z). So totally worth the need to authenticate.

I'd still keep the CDS module for backwards compatibility. The EDH mirror is the same as the NCAR mirror in that it stores the data already regridded at 0.25/0.25 deg., so using it at any other resolution would yield results that are different than CDS. I don't know how many people actually use different grids, but I don't think they should start getting different results when they update `atlite`. 

I'd propose the following:
1. Add an `era5-edh` module. Either raise a Warning if people try to use resolutions other than 0.25/0.25, or just get rid of interpolation code altogether and throw an Exception
2. Add a warning that `era5` will be deprecated in a future releases.
3. At some point either silently switch over `era5-edh` to `era5` or do a breaking change.

What do you think? 

I'll look into swapping the NCAR code with EDH code and update this PR.

fneum:
Great, I think the restriction to 0.25 deg downloads is totally fine. Users could always reinterpolate if they need to. Let's start with `era5-edh` module, keep it in parallel with `era5` and make some experience first before deprecating CDSAPI module.

mrmorawski:
I implemented the EDH source as discussed above @fneum.

Some performance stats (this time on a 6mo Europe cutout):

  | Metric            | CDS (`era5`)        | EDH (`era5-edh`) |
  |-------------------|---------------------|------------------------|
  | Wall clock (dl + process) | 5478 s (91 min) | 1024 s (17 min)    |
  | Memory peak       | 1.87 GB             | 5.91 GB                |
  | Network rx        | 3.79 GB             | 6.99 GB                |
  | Cutout size       | 4.71 GB             | 3.43 GB                |

There are minor numerical differences between the two sources, because EDH stores the dataset in a compressed form, but the results are not significant IMO. I documented the differences in [this notebook](https://github.com/mrmorawski/atlite/blob/era5-ncar-thredds/examples/comparing_era5_sources_cds_vs_edh.ipynb)

I profiled EDH by running it  large cutouts.  Memory usage stayed below 8GB for a 27GB cutout and below 10GB for a 44GB cutout.  To get those larger cutouts to work in a sensible amount of time, I needed to use `zstd` compression as mentioned in #508.

`era5-edh` is ready for review.

test/test_preparation_and_conversion.py
```diff
@@ -651,3 +652,143 @@ def test_all_non_na_gebco(cutout_gebco):
         Every cells should have data.
         """
         assert np.isfinite(cutout_gebco.data).all()
+
+
+class TestERA5EDH:
+    """
+    we assume that era5.py cutouts are correct, and only test whether era5_edh.py cutouts
+    have the same structure and contain the saem numerical values
```
fneum:
```suggestion
    have the same structure and contain the same numerical values
```

fneum (approved):
Looks really great! I played with it a bit locally as well. Thanks for the example as well!

One question is why the memory peak is so different.

fneum:
@mrmorawski, let me know when I should give this another review regarding the latest changes.

mrmorawski:
Thanks for checking in @fneum !

I've been profiling and optimising memory usage, I think I mostly have it figured out. I will give you a heads up soon, most likely this week.

-------- Original Message --------

mrmorawski:
## Profiling memory usage

I did a lot of memory profiling for `era5-edh`. I fixed some minor issues (chunk sizes, operation order), which brought the memory down to 3.5GB for the 6-month EU cutout I was measuring above. Here is memory usage over time:

image

I think I have identified a key bottleneck - system `malloc` not releasing small chunks of memory to the OS, which accumulate in the process. It's described in more detail [here](https://docs.coiled.io/blog/tackling-unmanaged-memory-with-dask.html#memory-trimming) , and seems to be Linux and MacOS specific. The `distributed` dask scheduler deals with this by launching workers with an environment variable `MALLOC_TRIM_THRESHOLD_=65536`, which causes `malloc` to release memory sooner and avoids accumulation. When I ran the same download as above with this env var, the peak memory usage dropped to 2.35GB.

image


## Memory usage conclusions

 - I think that for cutouts in the <50GB range memory usage is acceptable with the default settings.

- For cutouts above that, I think the users should use the `distributed` scheduler, as the default threaded one is not memory-aware and may OOM on smaller machines. I wrote an example on how to use the distributed scheduler [here](https://github.com/mrmorawski/atlite/blob/era5-ncar-thredds/examples/large_cutouts_with_era5_edh.ipynb) . It's two lines of code to give people granular control over memory usage, OOM prevention, nice graphs for monitoring progress etc. In the example notebook, I generate data for the whole world over a year while keeping memory below 12GB.

image


- Alternatively we could do the memory trimming in the code as done [here](https://docs.coiled.io/blog/tackling-unmanaged-memory-with-dask.html#memory-trimming) but IMO we should avoid calling `malloc` directly from Python if we can help it.


## Other issues (support for Python 3.10)
I also fixed issues flagged by the CI, now the tests pass (see [here](https://github.com/mrmorawski/atlite/actions/runs/27011619113/) except on Python 3.10. Python 3.10 doesn't work because we need Zarr-3, and 3.10 only supports Zarr-2. I couldy hack around this, but I think we should either:
- require Python>=3.11, or
- make `era5-edh` conditional on having Python>=3.11

Would dropping 3.10 be OK with you @fneum ? It'll [reach end-of-life in October 2026 anyway ](https://devguide.python.org/versions/)

mrmorawski:
I updated the branch to pass the new `ruff` rules checks introduced in https://github.com/PyPSA/atlite/commit/966380825089e5f913eb3dd45141abbff4fc075d

And one thing that I forgot to mention above - this branch introduces a new secret to the test CI (`EARTHDATAHUB_API_KEY`), which is needed to download `era5-edh` cutouts for tests. You'd need to add one to the GitHub Actions config in this repo.

fneum:
Yes, dropping Python 3.10 is fine (we did this in PyPSA already).

I'll add the earth data hub API key now.

mrmorawski:
Awesome. I see there are still some issues with the CI tests (they pass, but ERA5 and ERA5EDH are being skipped in tests for this PR). I will debug those next week.

Is there anything else you need to be able to merge this @fneum?

examples/create_cutout.ipynb
```diff
@@ -70,13 +69,14 @@
     "The `cutout` is created in the directory and file specified by the relative `path`\n",
     "If a cutout at the given location already exists, then this command will simply load the cutout again.\n",
     "If the cutout does not yet exist, it will specify the new to-be-created cutout.\n",
-    "> **Note** `ERA5`, Before the data can be downloaded it has to be processed by CDS servers, this might take a while depending on the ammout of data requested. \n",
+    "\n",
+    "> **Note** if using the `era5` module, data has to be processed by CDS servers, before it can be downloaded. This might take a while depending on the ammount of data requested. \n",
```
maurerle:
```suggestion
    "> **Note** if using the `era5` module, data has to be processed by CDS servers, before it can be downloaded. This might take a while depending on the amount of data requested. \n",
```

doc/installation.rst
```diff
@@ -40,6 +40,33 @@ If you're feeling adventurous, you can also install the latest master branch fro
...
+3. Click on the "i" icon next to a token in the "My personal access tokens" section
+4. Click on "Download .netrc". Save the resulting file in the project folder or in your home folder (or append to an existing ``.netrc``)
+   Alternatively, you can store the API key in this file as a ``EARTHDATAHUB_API_KEY`` environmental variable
+5. For Linux/Mac users, if using ``.netrc`` in your home folder, ensure that is has ``0600`` (owner read/write only) permissions
+   This doesn't apply if storing the file in your project folder
+
+Climate data store (fallback)
+-------------------------------
+
+1. Create an acocount at https://cds.climate.copernicus.eu/
```
maurerle:
```suggestion
1. Create an account at https://cds.climate.copernicus.eu/
```

atlite/resource.py
```diff
@@ -29,10 +29,9 @@
 CSPINSTALLATION_DIRECTORY = RESOURCE_DIRECTORY / "cspinstallation"
 
 if TYPE_CHECKING:
-    from typing import TypedDict
+    from typing import NotRequired, TypedDict
```
maurerle, line 32: I don't have anything against dropping 3.10 support, but I guess that 3.10 would work without this change?

Either remove `typing_extensions` from pyproject.toml as well, or leave this change as before (and remove python 3.10 in a separate PR)

maurerle (approved):
did take a look on this as well (just code though, did not try it). Sounds like a good alternative! :)

The history of this PR should be squashed or cleaned up though.

atlite/resource.py
mrmorawski: this PR needs Python 3.11+ to work because `era5-edh` relies on Zarr v3's `ObjectStore`, which is not available for Python 3.10 AFAIK. I could create a separate PR for 3.10->3.11 if you'd like that, but for now I just removed `typing_extensions `  from dependencies. Does this resolve your comment?

mrmorawski:
thanks for the review @maurerle !

I checked out the CI, and the reason that EDH tests were being skipped is that CI on PRs doesn't have access to secrets. CDS tests are not skipped, but only on Ubuntu, because there are the nightly built cached cutouts for those.

Regarding commit history, I totally agree it should be cleaned up, but I thought you'd just squash merge it @maurerle @fneum?

If you'd prefer that I squash this branch, I'd be happy to do that of course

fneum:
We just squash at the end.

mrmorawski:
Then I think it's ready for merge - CI runs all tests and [passes on my fork](https://github.com/mrmorawski/atlite/actions/runs/29029735415). The last remaining issue is the Python version bump - would you prefer a separate PR for that, or can we do it here?

maurerle (approved):
I think it is fine here.
I do not have commit rights, but I'd approve.

FabianHofmann:
hey @mrmorawski this is great stuff! I did some small refactoring and added a few tests:

- The wind and sun/radiation calculations were written out twice — once in the old CDS module and once in the new module; wrote them once in a shared spot.
- When the access key was wrong or expired, the test setup kept retrying for ~2 minutes and then said "the server didn't respond" — which wasn't true. Now it recognizes a bad key immediately and says so clearly.
- added a new test to check against unsupported grid. and one that the login-key lookup works 
- smaller things like removed comments or corrected docs (few old CDS-only options are quietly ignored by the new module)

mrmorawski:
Thanks for the review @FabianHofmann!

I agree that it's good to factor processing out, it will make it easier to fix #509. I didn't do it because I did not feel like I had the authority to touch a working module :)

However, I'm not sure about the EDH reachability function in `conftest.py`. The reason I had a catch-all retry is because otherwise tests were being skipped in the CI. When 9 runners simultaneously hit EDH with the same credentials, it sometimes starts rate limiting by responding with `429`s (see comment [here](https://github.com/mrmorawski/atlite/blob/3f45cc4359904d4c618aa13e8ae86710beeab155/test/conftest.py#L47)). Under your change, this will be treated as failure and EDH tests will be skipped. I see four options:

0. Keep the current code, sometimes EDH tests will skip, but it's no big deal
1. Revert your change to `_edh_reachable`, which costs up to 2min of waiting when there is an actual reachability/credential issue (rarely)
2.  `except` 429s, make `edh_reachable` a bit more complex
3. rework the CI pipeline, either by adding an intermediate test cutout build step, or by having per-runner EDH credentials. 

If you consider the wait too much, I'd recommend 2, I implemented it for now.

FabianHofmann:
Yeah I did not think about this case. Let's do as you propose :)

FabianHofmann (approved):
great contribution, thanks @mrmorawski

PR #513, opened 2026-07-09
feat: add circular raster buffer

jome1:
Add `buffer_geometry` argument to `ExclusionContainer.add_raster()`
Previously, the buffer applied to raster exclusions in `shape_availability()` always used scipy.ndimage.binary_dilation which produces a diamond-shaped (L1-metric) buffer: each iteration expands the excluded area by one pixel in the four cardinal directions, so the footprint after N iterations is a diamond of radius N.
This PR adds a `buffer_geometry` parameter to `add_raster()` that lets users choose between two buffer shapes:
	• "diamond" (default) – retains the existing binary_dilation behaviour. No change to existing code that does not pass buffer_geometry.
	• "circular" – uses `scipy.ndimage.distance_transform_edt` to compute exact Euclidean distances. Every pixel whose Euclidean distance to the nearest excluded pixel is ≤ buffer / res is marked as excluded, producing a geometrically accurate circular buffer.

According to Claude, the circular buffer method migth also be faster:
"Performance: The circular method runs in O(pixels) with a single linear-time scan regardless of buffer size. The diamond method runs in O(N × pixels) where N = buffer / res, meaning it becomes proportionally slower as the buffer grows. For typical use cases (e.g. a 500 m buffer at 100 m resolution, N = 5) the circular method is already several times faster, and the gap widens with larger buffers."

## Additional info
I got aware of the raster buffering while testing atlite raster exclusion with buffering. I measured the distance of the buffer in the output raster and the buffer distance was only correct along the grid (vertically and horizontally) but not in diagonal directions.
I used Claude to understand the current raster buffering and to implement the circular buffering. 
Before adding realease notes and add the argument to `add_raster()` in the documentation, please give the PR a review and let me know if it makes sense.

I used the following code in a Jupyter Notebook (separate code cells) to test the results (Let me know if you need the input files to test it)

```python
import atlite
import os
import geopandas as gpd
import matplotlib.pyplot as plt

regionPath = os.path.join("Odense_square_EPSG4326.geojson")
region = gpd.read_file(regionPath)
region = region.to_crs(25832) #local UTM zone
print(region.crs)

# initiate Exclusion container
excluder = atlite.ExclusionContainer(crs=25832, res=10)
# add raster (single pixel) with buffer
excluder.add_raster("single_pixel.tif", codes=1, buffer=500, buffer_geometry="diamond", crs=4326)
masked, transform = excluder.compute_shape_availability(region)
# plot
fig, ax = plt.subplots()
excluder.plot_shape_availability(region)

# initiate Exclusion container
excluder = atlite.ExclusionContainer(crs=25832, res=10)
# add raster (single pixel) with buffer
excluder.add_raster("single_pixel.tif", codes=1, buffer=500, buffer_geometry="circular", crs=4326)
masked, transform = excluder.compute_shape_availability(region)
# plot
fig, ax = plt.subplots()
excluder.plot_shape_availability(region)
```

These are the plots with the different raster buffering methods (there is a single pixel in the center. The buffer is applied around this pixel):

image

image

## Checklist

### DIFF
--- a/atlite/gis.py
+++ b/atlite/gis.py
@@ -601,20 +601,20 @@ class ExclusionContainer:
             Use this to create a buffer around the excluded/included area.
             The default is 0.
         buffer_geometry : {"diamond", "circular"}, optional
-            Shape of the buffer kernel applied to raster exclusions:
+            Shape of the buffer applied around raster exclusions. Only used
+            when ``buffer > 0``. The default is ``"diamond"``.
 
-            * ``"diamond"`` (default) – uses :func:`scipy.ndimage.binary_dilation`
-              with the default 4-connected cross structuring element, which
-              produces a diamond (L1-metric) footprint. Equivalent to the
-              historic behaviour.
-            * ``"circular"`` – uses :func:`scipy.ndimage.distance_transform_edt`
-              to compute exact Euclidean distances, giving a true circular
-              (L2-metric) buffer. This is both more geometrically accurate and
-              significantly faster for large buffers because it runs in O(pixels)
-              regardless of buffer size, whereas the diamond approach scales as
-              O(buffer/res × pixels).
-
-            Only used when ``buffer > 0``.
+            * ``"diamond"`` – dilates the mask with
+              :func:`scipy.ndimage.binary_dilation`, expanding it by one pixel
+              in the four cardinal directions per iteration. The footprint is a
+              diamond (L1 metric), so the buffer distance is only accurate along
+              the grid axes and overreaches diagonally. This is the historic
+              behaviour.
+            * ``"circular"`` – marks every pixel whose exact Euclidean distance
+              to the mask (via :func:`scipy.ndimage.distance_transform_edt`) is
+              within ``buffer``, giving a geometrically accurate circular
+              (L2 metric) buffer. Prefer this when diagonal accuracy matters;
+              note that in practice it is typically slower than ``"diamond"``.
         nodata : int, optional
             Value to use for nodata pixels. The default is 255.
         invert : bool, optional
--- a/atlite/gis.py
+++ b/atlite/gis.py
@@ -27,6 +27,7 @@ from rasterio.mask import mask
 from rasterio.plot import show
 from rasterio.warp import transform_bounds
 from scipy.ndimage import binary_dilation as dilation
+from scipy.ndimage import distance_transform_edt
 from shapely.ops import transform
 from shapely.strtree import STRtree
 from tqdm import tqdm
@@ -462,8 +463,12 @@ def shape_availability(
         if d["invert"]:
             masked_ = ~masked_
         if d["buffer"]:
-            iterations = int(d["buffer"] / excluder.res) + 1
-            masked_ = dilation(masked_, iterations=iterations)
+            if d["buffer_geometry"] == "circular":
+                radius = d["buffer"] / excluder.res
+                masked_ = distance_transform_edt(~masked_) <= radius
+            elif d["buffer_geometry"] == "diamond":
+                iterations = int(d["buffer"] / excluder.res) + 1
+                masked_ = dilation(masked_, iterations=iterations)
 
         exclusions = exclusions | masked_
 
@@ -557,6 +562,7 @@ class ExclusionContainer:
         | Callable[[NDArray], NDArray]
         | None = None,
         buffer: float = 0,
+        buffer_geometry: str = "diamond",
         invert: bool = False,
         nodata: int = 255,
         allow_no_overlap: bool = False,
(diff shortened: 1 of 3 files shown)


atlite/gis.py
```diff
@@ -462,8 +463,12 @@ def shape_availability(
...
             masked_ = ~masked_
         if d["buffer"]:
-            iterations = int(d["buffer"] / excluder.res) + 1
-            masked_ = dilation(masked_, iterations=iterations)
+            if d["buffer_geometry"] == "circular":
+                radius = d["buffer"] / excluder.res
+                masked_ = distance_transform_edt(~masked_) <= radius
+            elif d["buffer_geometry"] == "diamond":
+                iterations = int(d["buffer"] / excluder.res) + 1
+                masked_ = dilation(masked_, iterations=iterations)
```
maurerle, line 481: A base case is missing here, instead of silently doing nothing.

Also importing old buffers would not have an entry for buffer_geometry, and do nothing.
This would break importing older data. Adding a default to "diamond", if nothing is set would be necessary

maurerle (changes requested):
Great idea. This looks fine in general.

I made a suggestion to improve this here:
https://github.com/jome1/atlite/pull/1

and also added unit tests to verify the suggested improvements.

Regarding the performance claim, it looks as if circular would be faster, because it only does one (heavy) iteration, instead of multiple iterations.
A small test did show that diamond is still much faster:

```
Generating 5000x5000 mask with a diagonal exclusion line...
Buffer Size (px)   | Diamond Time (s)   | Circular Time (s)  | Speedup Ratio
---------------------------------------------------------------------------
20                 | 0.32025            | 2.02672            | 0.16
100                | 0.33293            | 2.01348            | 0.17
1000               | 0.50594            | 2.00961            | 0.25
10000              | 0.83164            | 1.98676            | 0.42
100000             | 0.82718            | 2.03129            | 0.41
1000000            | 0.82630            | 2.02098            | 0.41
```

Just as an info aside

<details>
<summary>benchmark source code</summary>

```python
import time
import numpy as np
import scipy.ndimage

def run_benchmark():
    # Large 5000 x 5000 grid (25 million pixels)
    shape = (5000, 5000)
    print(f"Generating {shape[0]}x{shape[1]} mask with a diagonal exclusion line...")

    mask = np.zeros(shape, dtype=bool)
    # Add a diagonal line representing a linear exclusion (e.g. road, river, or coastline)
    for i in range(shape[0]):
        mask[i, i] = True

    buffer_sizes = [20, 100, 1000, 10000, 100000, 1000000]

    print(f"{'Buffer Size (px)':<18} | {'Diamond Time (s)':<18} | {'Circular Time (s)':<18} | {'Speedup Ratio':<15}")
    print("-" * 75)

    for N in buffer_sizes:
        # Benchmark Diamond Buffering (L1 Metric / binary dilation)
        t0 = time.perf_counter()
        res_diamond = scipy.ndimage.binary_dilation(mask, iterations=N)
        t_diamond = time.perf_counter() - t0

        # Benchmark Circular Buffering (L2 Metric / EDT)
        t0 = time.perf_counter()
        if mask.any():
            res_circular = scipy.ndimage.distance_transform_edt(~mask) <= N
        else:
            res_circular = mask
        t_circular = time.perf_counter() - t0

        speedup = t_diamond / t_circular
        print(f"{N:<18} | {t_diamond:<18.5f} | {t_circular:<18.5f} | {speedup:<15.2f}x")

if __name__ == "__main__":
    run_benchmark()
```

</details>

atlite/gis.py
jome1: Ah yes, very good! 
"diamond" is also the default in the `add_raster()` function, so old code does not need to be updated with the buffer_geometry argument.

jome1: This is resolved in your PR with the raising the ValueError right? Or do you want me to adjust something?

jome1:
Amazing! Thanks for you swift review and the improvements @maurerle!
I merged your PR.

Great that you did an actual speed test. Just too bad that the "circular" buffer is much slower. I am running a big resource potential study for Kenya at 10m resolution. I have not test the "circular" buffer yet, but so far the runs always needed around 2h. I want to use the "circular" buffer because it makes more sense, but then it will take even longer...

maurerle:
@jome1 I tested on a different computer where the differences between the two was much lower.
YMMV so I'd recommend you to run the provided benchmark script (in my comment above) on your hardware.

test/test_gis.py
```diff
@@ -646,3 +646,66 @@ def test_plot_shape_availability(ref, raster):
     with warnings.catch_warnings():
         warnings.simplefilter("ignore")
         excluder.plot_shape_availability(shapes)
+
+
+def test_exclusion_container_serialization_backward_compatibility(ref, raster):
+    """Test pickle serialization of ExclusionContainer and backward compatibility with older configurations."""
```
FabianHofmann: I would not talk about backwards compat here. wording should be in absolute terms (what is this function testing)

FabianHofmann (approved):
thanks for the pr. this makes very much sense. I took the freedom to improve the doc strings and the test setup. while I was on it I added the release note. please take a look at the last commit. after confirming this is ready to go. I would argue we should add a follow up issue/pr to pave the way to a make "circular" the new default

jome1:
Thanks for the review @FabianHofmann! 
I confirm. Do you have to merge? I think I cannot. There is still "1 requested change" by @maurerle, but I did merge his PR...

Do you want me to open another PR to make "circular" the new default?

FabianHofmann:
Yes, I can merge after the CI is through. Not sure when to change the default. before deciding that I would like to see it in action first and perhaps we can do some benchmark with the pypsa-eur code? if things look fine, we can take this to the new release

PR #519, opened 2026-09-01
Fix turbine_type search parameter in get_oedb_windturbineconfig

dylanpulver:

`get_oedb_windturbineconfig` documents `turbine_type` as a recognized search
argument, but the branch that applies it reads the wrong key:

```python
if "turbine_type" in search_params:
    selector &= df.turbine_type.str.contains(search_params["name"], case=False)
```

Two consequences on current `master`, both reproduced against the live OEDB
`wind_turbine_library` endpoint the function queries:

* `get_oedb_windturbineconfig(turbine_type="E-101/3050")` raises
  `KeyError: 'name'`.
* `get_oedb_windturbineconfig(name="E-101/3050", turbine_type="__nope__")`
  returns `E-101/3050`. The turbine type is silently ignored rather than
  narrowing the search.

The two columns are not interchangeable. Of the 68 library entries that carry
a power curve, 58 have `name != turbine_type`, 51 have a `turbine_type` that
is not a substring of its own `name`, and 7 have no `name` at all (several
Nordex rows), so for those the turbine type is the only string that
identifies them. `manufacturer` and `name` are unaffected and are used as
controls in the new tests.

The fix passes `search_params["turbine_type"]`. Introduced in `12846ce`
(v0.2 preparation, Jan 2021).

The three new tests patch the module-level `_oedb_turbines` cache with a
two-row frame, so they are deterministic and make no network call. Reverting
the fix fails the first two; filtering the right value against the `name`
column fails the first; deleting the branch fails both. `test_resource.py`
previously asserted only that a result was truthy and never passed
`turbine_type=`, which is why the defect survived.

Test suite: 83 passed / 38 skipped on `master` → 86 / 38 with this branch,
the +3 being exactly the new tests. The 38 skips are the CDS-credentialed
cutout tests and are skipped identically either way, so that path is unverified
here. `ruff check` and `ruff format --check` are
clean across the repo.

Written with AI assistance; the behaviour above was measured, not inferred.

## Checklist

### DIFF
--- a/RELEASE_NOTES.rst
+++ b/RELEASE_NOTES.rst
@@ -35,6 +35,11 @@ Upcoming Release
 
 **Bug fixes**
 
+* Fix ``get_oedb_windturbineconfig`` applying the documented ``turbine_type``
+  search parameter to the value of ``name``. Searching by ``turbine_type``
+  alone raised ``KeyError: 'name'``, and combining it with ``name`` silently
+  ignored the requested turbine type.
+
 * Fix ``Cutout.line_rating`` passing line azimuth in radians while
   ``convert_line_rating`` interpreted ``psi`` as degrees. Azimuths are now
   computed in degrees, matching the documented unit.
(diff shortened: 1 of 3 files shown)


FabianHofmann (approved):
great fix. thank you @dylanpulver

